PR #15, opened 2019-06-27
Clear advanced filters on visibility toggle

### LINKED ISSUE #13: Clear values from redux when filters are hidden
Right now there is a `More filters` link on the results page that will expand some additional "advanced" filters. When they are shown or hidden, they are unregistered from `redux-form` which is the expected behavior, but the values are not getting cleared from redux. So someone could potentially select a few options but then collapse the advanced filters and those hidden values will still get passed through to the API. I think the ideal behavior would be to empty out those values if they are hidden.

el-mapache:
Fixes #13 

* Only clears language for now
* Adds filter actions, advanced filter plugin to redux-form reducer, and
dispatch method to search component
* Updates package.json with `dev` script that docker-compose file expects

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
     "tailwindcss": "^1.0.1"
   },
   "scripts": {
+    "dev": "npm install && npm run start",
     "start": "npm run tailwind:css && react-scripts start",
     "build": "npm run tailwind:css && react-scripts build",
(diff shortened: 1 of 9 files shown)


el-mapache:
Opening a WIP for visibility while I add tests!

Issue #17, opened 2019-06-27
Add new filters to type of care


allisonnorman:
Can you please implement two new filter options in the Type of Care section? 

The default option should still be All providers, but we'd like to add two new ones beneath that:

- Add "Intake assessment offered" (No service code, returns ALL results, just like All Providers now) - this is to get a sense of whether the integrated intake assessment content is more effective or whether visitors gravitate to this option instead
- Add "Interim services for clients available" (we may change that name later but I can do it) (Service code: ISC)

Order should be:
All providers
Intake assessment offered
Interim services for clients available
Detox
.... (as it stands today)

allisonnorman:
Adding high priority label because these updates will complete our initial pass at filtering and once in place, we can start incorporating the integrated content components

Issue #14, opened 2019-06-27
Refactor use of URLSearchParams in buildParams function


hursey013:
This is not supported by IE, so need to find a polyfill or alternative.

PR #24, opened 2019-06-28
Remove URLSearchParams

hursey013:
Fixes https://github.com/18F/samhsa-prototype/issues/14
Fixes https://github.com/18F/samhsa-prototype/issues/17

Refactors the way the params are prepared to be sent to the endpoint, no longer relying on URLSearchParams which is not cross browser compatible.

Also adds additional options to the `Type of Care` filter.

### DIFF
--- a/src/actions/facilities.js
+++ b/src/actions/facilities.js
@@ -1,3 +1,4 @@
+import qs from 'qs';
 import API, { buildParams } from '../utils/api';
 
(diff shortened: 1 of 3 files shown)


hursey013:
@el-mapache definitely would like to take another stab at it if you're up for pairing next week.

allisonnorman:
The type of care changes look great! Thanks @hursey013!

PR #27, opened 2019-06-28
Add button to reset filters on ResultsList page

### LINKED ISSUE #9: Consider supporting a way to reset filters
I'm wondering if it makes sense to allow users to reset filters (remove applied filters) on the results page. I think this could help when people have no or limited results, or want to refine the results differently given that we're asking them to select specific filters one at a time.

It could be as simple as a reset link near the top or by the update filters button. If others agree, we can also get design feedback. Attached is an example from Redfin for reference. 
Screen Shot 2019-06-25 at 3 09 01 PM

el-mapache:
Fixes #9

**Handy gif**

filter-reset

### DIFF
--- a/src/components/Form/Filters.js
+++ b/src/components/Form/Filters.js
@@ -48,9 +48,9 @@ export class Filters extends Component {
 
     return (
       <form onSubmit={handleSubmit} css={tw`mb-6`}>
-        <RowFlex>
-          <h2 css={tw`w-5/6 mb-6`}>Filters</h2>
-          <div css={tw`w-1/6 h-12`}>
+        <RowFlex css={tw`justify-between`}>
+          <h2 css={tw`px-3`}>Filters</h2>
+          <div css={tw`px-3`}>
             <button
               className="reset-filters"
               css={tw`inline-block mt-1 font-bold text-blue-500 hover:text-blue-800`}
--- a/src/components/Form/Filters.js
+++ b/src/components/Form/Filters.js
@@ -1,15 +1,15 @@
 import React, { Component } from 'react';
-import { Field, reduxForm } from 'redux-form';
+import { Field, reduxForm, getFormValues } from 'redux-form';
 import styled from 'styled-components/macro';
 import tw from 'tailwind.macro';
 import { connect } from 'react-redux';
-import { getFormValues } from 'redux-form';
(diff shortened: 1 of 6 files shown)


el-mapache:
open question: does resetting filters also clear the search results?

el-mapache:
@hursey013 This is ready to go. I'm going to make a follow up issue to refactor the `location` field into it's own slice of form state. The code is getting a little tricky to treat `location` separately from the other filters.

src/components/Form/Filters.js
```diff
@@ -37,12 +38,29 @@ export class Filters extends Component {
     );
   };
 
+  handleReset = () => {
+    this.props.resetAllFilters({ location: this.props.location });
```
hursey013: Does `{ location: this.props.location }` need to be passed in here?

src/components/Form/Filters.js
```diff
@@ -37,12 +38,29 @@ export class Filters extends Component {
...
+  };
+
   render() {
     const { handleSubmit, data } = this.props;
     const { isHidden } = this.state;
 
     return (
       <form onSubmit={handleSubmit} css={tw`mb-6`}>
+        <RowFlex>
+          <h2 css={tw`w-5/6 mb-6`}>Filters</h2>
```
hursey013: There's some sort of alignment issue going on with this row, you ok if I add a commit to fix it up?

src/components/Input/Location.js
```diff
@@ -33,6 +33,7 @@ class Location extends Component {
             queryDelay={100}
             onFocus={this.handleFocus}
             onSuggestSelect={this.handleSuggest}
+            onChange={value => input.onChange(value)}
```
hursey013: Do you mind explaining what this is doing? Is this to support the validation?

src/components/Form/Filters.js
el-mapache: yes, because the `handleReceiveFacilities` method needs to know what the current location is to perform a search.

src/components/Form/Filters.js
el-mapache: go for it

src/components/Input/Location.js
el-mapache: This was a hold-over from a previous iteration where the location was cleared along with the other filters. Removing!

PR #35, opened 2019-07-01
Prevent searches with invalid location data

### LINKED ISSUE #29: Investigate no results error when I search and submit a zip quickly
Sometimes when I type a zip code in quickly on location search, I get an oops, no results message.

After debugging with @hursey013, we think it's because of the short delay that is applied, and the api request hasn't completed by the time I submit.

el-mapache:
Fixes #29 

* Disables search button on homepage when an invalid location is present
* Adds <Button> styled component
* Automatically select the first valid suggestion when a user blurs the
location input field.
* Adds <Homepage> component-level submit handler to check for valid location
prop — this is done so that we can avoid using the native `disabled` property
on the submit button. Using that property breaks tab order and may negatively
impact users who rely on assistive devices

### DIFF
--- a/src/components/Input/Location.js
+++ b/src/components/Input/Location.js
@@ -29,7 +29,6 @@ class Location extends Component {
             types={['(regions)']}
             placeDetailFields={[]}
             autoActivateFirstSuggest={true}
-            queryDelay={0}
             initialValue={input.value.label}
             queryDelay={100}
             onFocus={this.handleFocus}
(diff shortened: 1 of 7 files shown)


el-mapache:
Sorry for the delay, I believe this is ready to go.

src/components/Input/Location.js
```diff
@@ -22,14 +31,9 @@ class Location extends Component {
             autoActivateFirstSuggest={true}
             queryDelay={0}
```
hursey013: I snuck this in here so it looks like we've got two `queryDelay` props, so we can get rid of this `0` instance.

el-mapache:
@hursey013 Thanks! I'll remove the extra prop and merge.

PR #37, opened 2019-07-01
Handle 100+ distance properly in new api reducer

hursey013:
The `All` value for distance was falling through the reducer and getting added to `sCodes`, this makes sure it is handled properly.


hursey013:
Are you alright if I merge this in so it's fixed for the usability test later today and then follow up in another PR with a test?

el-mapache:
@hursey013 yep, go for it!

PR #38, opened 2019-07-01
Update filter names

allisonnorman:
Updating filter names based on @amandacostello's content doc: https://docs.google.com/document/d/1gnoON2VLdI5HBVG1dcghdfedzwuj1KJed62b1-71Oz8/edit#

### DIFF
--- a/src/utils/filters.js
+++ b/src/utils/filters.js
@@ -1,6 +1,6 @@
 export const age = [
-  { value: 'ADLT', label: 'Adults' },
-  { value: 'CHLD', label: 'Children/adolescents' }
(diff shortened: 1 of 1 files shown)


allisonnorman:
Couple outstanding questions to @amandacostello before I merge:
- I noticed Assessment/Pretreatment has initial caps on Pre, but Children/adolescents doesn't on adolescents. Which should we use for consistency?
- Also, Sliding fee scale (payment assistance where fee is based on income and other factors)  makes the filter section pretttty wide. Any thoughts on shortening that?

PR #39, opened 2019-07-01
Updating filter labels

allisonnorman:


allisonnorman:
Checking with @amandacostello on the length of the new label before merging

PR #40, opened 2019-07-01
Adding new filter option under type of care

allisonnorman:
We are going to add one MH-specific filter option to type of care to get a sense of the MH engagement. 

(NOTE: I am not sure if I can just add this so going to create a branch to check before merging)


allisonnorman:
@hursey013 - This seems to have worked, which means your code is amazing. Would you mind reviewing just to make sure anything else needs to be done when adding a new filter option?

PR #45, opened 2019-07-02
Adds tests to API query reducer

hursey013:
Tests the various conditions with the `buildParams` reducer.


src/utils/api.js
```diff
@@ -9,19 +9,20 @@ export const buildParams = query => {
   const initialValues = {
```
el-mapache: This wasn't added in this PR, but we can move `initialValues` out of the function. Otherwise we are reinitializing this object every time the function is called.

PR #51, opened 2019-07-03
Add content page template

hursey013:
Adds a new view for content based pages. As it stands now, the content is pulled out into `utils/topics.js` - and if the `body` of a `subTopic` is empty its just filled with lorem ipsum. The way all of this works will need to be refined so that we can support things like inline formatting and links within the content, but this helps gets some of the pieces in place.

I also linked up the homepage cards to the corresponding pages for the time being.

Preview: https://cg-efcd7c90-b365-4336-8b87-6174f1975c99.app.cloud.gov/preview/18f/samhsa-prototype/bh-content-page/

### DIFF
--- a/src/components/Page.js
+++ b/src/components/Page.js
@@ -1,0 +1,66 @@
+import React, { Component } from 'react';
+import { withRouter, Link } from 'react-router-dom';
+import { HashLink } from 'react-router-hash-link';
+import 'styled-components/macro';
+import tw from 'tailwind.macro';
+import NoMatch from './NoMatch';
+import { convertToSlug } from '../utils/misc';
+import { topics, lipsum } from '../utils/topics';
+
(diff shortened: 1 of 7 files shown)


src/components/Page.js
```diff
@@ -0,0 +1,66 @@
...
+import 'styled-components/macro';
+import tw from 'tailwind.macro';
+import NoMatch from './NoMatch';
+import { convertToSlug } from '../utils/misc';
+import { topics, lipsum } from '../utils/topics';
+
+class Page extends Component {
+  render() {
+    const { match } = this.props;
+    const topic = topics.find(({ id }) => id === match.params.pageId);
```
el-mapache: What do you think about moving this logic into `componentDidMount` and redirecting to the 404 page instead?

PR #52, opened 2019-07-05
Update homepage and nav labels

allisonnorman:
Changed to "Quickly find providers who treat substance use disorders and addiction"


allisonnorman:
@Jkrzy or @el-mapache - Would you mind reviewing this on Monday? Fairly confident these are straight-forward changes but would be great to get another look before I push to master. Thanks!

PR #54, opened 2019-07-05
Changed no results messaging

allisonnorman:
- Removed blue button on no results messaging
- Changed text per Amanda's content doc


allisonnorman:
Tried adjusting the no results messaging in [#55](https://github.com/18F/samhsa-prototype/issues/55). I think we need @amandacostello to take a look and adjust as it doesn't feel right as it stands.

allisonnorman:
Note: break tag looks okay in mobile view but let me know if I should create an issue to do this better in the future.

Issue #53, opened 2019-07-05
Add integrated content text on home and results pages


allisonnorman:
1. On the homepage, below the location search box (inline with the location label and text box), add a subtitle (smaller font): "All facilities are licensed by their states. Provider data is collected, maintained, and verified by the Substance Abuse and Mental Health Services Administration (SAMHSA) through an annual survey."

2. On the results page, above facility cards and under the label Results (inline with results, smaller font), add: 
"Before visiting a facility, call to make sure they have the services you need. Not sure what you need? All facilities are licensed by their states, and provide assessments."

3. On the results page, underneath filters, add a subtitle (smaller font): "Filter the search results for facilities that match your needs." 

4. On the results page, underneath Update providers button, add a subtitle (smaller font): "Too many or too few results? Add or remove search filters related to the treatment you’re looking for." _(Note: on the attachment, I split this into two lines but feel free to do what looks best)_

[Here's a google doc](https://docs.google.com/document/d/1WrdvO1w0210EexY84kotiG81JNkutk7Qz5eCkg_IDck/edit) with screen shots to illustrate placement. Spacing is hard to show on my mock up but the subtitles should be smaller text and we should maintain the same kind of spacing as we have already once the text is added. 

Text is in red to highlight but should be black on the prototype.

PR #63, opened 2019-07-10
Add integrated content

hursey013:
Fixes https://github.com/18F/samhsa-prototype/issues/53


allisonnorman:
LGTM! Won't merge because checks are still underway but we're good to go once those are green.

Issue #59, opened 2019-07-09
Add beta testing banner with link to recruitment form


allisonnorman:
"**We are testing a new design for our treatment locator. Have feedback? We want to hear from you (hyperlink: https://forms.gle/35ZHQCGBkxCsJcs78).**
Visit current site at findtreatment.samhsa.gov (hyperlinked)

Some thoughts on placement: a banner above the US government website banner? Bold color (yellow?) similar to digital.gov. Happy to chat about this to see if you have other ideas.

PR #64, opened 2019-07-10
Adds beta banner

hursey013:
Fixes https://github.com/18F/samhsa-prototype/issues/59


allisonnorman:
Looks great! I noticed you didn't include the link back to the existing TL. Did it look odd in context? I was thinking of it being on a second line, less prominent/not bold but maybe it is confusing linking to two different sites (the form and TL)?

hursey013:
@allisonnorman I missed the text in the issue - just added it in.

PR #68, opened 2019-07-10
First pass at adding in info content

allisonnorman:
- Updated subtopics for 'treatment options'
- Added content to 'what happens next' and 'types of providers'

@hursey013 - It seems like the styling is getting stripped away, especially with spacing between p tags and the lists, which I'm sure is a limitation you mentioned before I got started but I'm wondering how we can tackle this given the paragraphs and headers. Before I go too much further down this path, I wanted to check to see if there is a better way for me to be inputting this.


hursey013:
@allisonnorman everything looks great so far! I've added some base styles for the content so you should see proper spacing for paragraphs and lists now.

hursey013:
@allisonnorman is this still a `don't merge`? I say lets get this merged in so that we can iterate off this content in the main branch.

allisonnorman:
No, it's good to go now. I'll merge.

PR #71, opened 2019-07-11
Refactor details component to pull from redux store

hursey013:
Derive the details page from passing in the `frid` and pulling the associated details from the store.

### DIFF
--- a/src/components/Card.js
+++ b/src/components/Card.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import { PropTypes } from 'prop-types';
 import { Link } from 'react-router-dom';
 import 'styled-components/macro';
 import tw from 'tailwind.macro';
@@ -8,6 +9,7 @@ import Button from './Form/Button';
 
 const Card = props => {
   const {
+    frid,
     name1,
     name2,
     miles,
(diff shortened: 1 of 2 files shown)


src/components/Card.js
```diff
@@ -77,4 +79,19 @@ const Card = props => {
   );
 };
 
+Card.propTypes = {
```
el-mapache: I like to add these to the top of the file, then devs who are working with it can see right up front what data the component requires.

src/components/Details.js
```diff
@@ -68,25 +76,26 @@ class Details extends Component {
...
-                <ul
-                  css={tw`text-sm leading-relaxed text-gray-700 list-disc list-inside`}
-                >
-                  {service.f3.split('; ').map((item, index) => (
-                    <li key={index}>{item}</li>
-                  ))}
-                </ul>
-              </div>
-            ))}
+            {services &&
```
el-mapache: In general I'm not super hot on having a bunch of inline anonymous functions in the render method, since react will have to recreate these every time a render is performed. Especially when you have larger chunks of DOM like this I find it a bit nicer to break them out into their own functions or components. 

I don't know that you will necessarily get a performance hit with this pattern, but I think it is a little less clean.

PR #75, opened 2019-07-12
Use build environment to determine API and GA Tracking ID

Jkrzy:
Leveraging Federalist's build system to configure the API URL and GA Tracking ID of deployed instances of the site based on their `branch`


src/components/Details.js
```diff
@@ -30,12 +31,21 @@ class Details extends Component {
     );
   }
 
+  flagData = frid => {
```
hursey013: Do you have any thoughts on how to reflect this back to the user? Alert window, or maybe change the text of the button and disable it?

.env.development
```diff
@@ -1,2 +1,2 @@
-REACT_APP_API_URL=http://localhost:9011/locator/
-REACT_APP_GA_TRACKING_ID=UA-143220884-2
+REACT_APP_DEV_API_URL=http://localhost:9011/locator/
+REACT_APP_BRANCH=dev
```
hursey013: Is this necessary? If we leave it off it should evaluate to false in the conditional statements below.

Jkrzy: Good catch, thanks!

PR #94, opened 2019-07-16
Reset filter location on subsequent homepage searches

hursey013:
Fixes bug with location field displaying old data.

To recreate: Search for a location on homepage. Get results. Hit back to go to homepage. Search for a new location, see that results are displayed accurately, however the old location is still present in filter box.

### DIFF
--- a/src/plugins/filters.js
+++ b/src/plugins/filters.js
@@ -1,4 +1,8 @@
-import { RESET_ADVANCED_FILTERS, RESET_ALL_FILTERS } from '../actions/filters';
+import {
+  RESET_ADVANCED_FILTERS,
+  RESET_ALL_FILTERS,
+  DESTROY_FILTERS
+} from '../actions/filters';
 
 export const initialFilterState = {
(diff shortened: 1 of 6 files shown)


src/plugins/filters.js
```diff
@@ -33,6 +37,9 @@ export default {
           values: { ...initialFilterState, location: state.values.location }
         };
       }
+      case DESTROY_FILTERS: {
+        return undefined;
```
el-mapache: Do we need to preserve the distance filter here? If so I'd recommend returning the initial state rather than `undefined`.

hursey013: Setting it to undefined is what was recommended in the docs: https://redux-form.com/8.2.2/docs/faq/howtoclear.md/#a-you-can-use-a-href-api-reducerplugin-md-the-code-plugin-code-api-a-to-teach-the-code-redux-form-code-reducer-to-respond-to-the-action-dispatched-when-your-submission-succeeds-

Issue #90, opened 2019-07-15
"Loading.." briefly appears when it should not


allisonnorman:
From feedback session: The word “Loading…” briefly appears in place of the location input field when initially loading

PR #96, opened 2019-07-16
Remove loading message from location input field

hursey013:
Fixes https://github.com/18F/samhsa-prototype/issues/90


allisonnorman:
I haven't been able to reproduce this in production. Any tips so that I can verify? Besides having a faster reaction time? :)

hursey013:
@allisonnorman If you go into chrome developer tools > Network > and under throttling select "Slow 3G" you should be able to slow things down enough to see it briefly.

PR #98, opened 2019-07-16
Updating SAMHSA description on home page

allisonnorman:
Updated language under search bar re: trustworthiness/source of data


allisonnorman:
@hursey013 - Not sure why this is failing?

PR #104, opened 2019-07-17
Add polyfill

hursey013:
Adds support for css position:sticky in IE11.


allisonnorman:
Hey @hursey013 - The info content sections preview links for this blank. Is that the best way to check this or should I be looking at something else?

hursey013:
No that's not right, let me check.

hursey013:
I don't love this implementation, going to close it in favor of another approach.

Issue #92, opened 2019-07-15
Add styling on info pages


allisonnorman:
Add more line breaks between sections and paragraphs on the info content pages.

From Nick: Consider adding some space below the subhead to match the spacing between each section (current spacing indicated with red boxes)
![Screen Shot 2019-07-15 at 4 58 59 PM](https://user-images.githubusercontent.com/6900926/61256508-db81ff80-a721-11e9-960d-c028a5df7487.png)

PR #108, opened 2019-07-17
Iterate typography on info pages

hursey013:
Fixes https://github.com/18F/samhsa-prototype/issues/92

- Increases spacing and adds dividers between topics
- Limits line length for better readability
- Adjust size and margin of headings, paragraphs, and lists


allisonnorman:
Afraid to resolve the branch conflict myself so are you able to, @hursey013?

Issue #109, opened 2019-07-17
Update What to expect - "home page" of info content (lede + content)


allisonnorman:
From @amandacostello: 

(Covered in PR #110)

- [x] Update homepage under "What to Expect" - "Treatment works, help is available, and people recover every day."
- [x] Update "What to Expect" landing page: Lede: Treatment works, help is available, and people recover every day."
- [x] Update Intro para: "Beginning recovery can be overwhelming and intimidating. While everyone's path to recovery is unique, there are certain things you can expect and be ready for when you start.
- [x] Update secondary para: There is help. From understanding more about addiction and treatment options, to being ready to navigate insurance and payment, this information can help you feel more confident as you take the next step to finding treatment.
- [x] Add lede/subtitle on homepage under What to expect

We should also:
- [x] Link to the Getting started content section in top nav
- [x] Remove subtopic from Getting Started section (or, if easier, rewrite that so it's not duplicative)

hursey013:
Closed with https://github.com/18F/samhsa-prototype/pull/110

PR #110, opened 2019-07-17
Updated 'What to expect' and other content changes

allisonnorman:
- Added 'what to expect' content from Amanda 
- Updated home page ledes
- Added subtitle under "what to expect" on homepage (I asked Nick to provide more feedback on this as part of his work on that section)

Still to do, which might be better for @hursey013 to do: 
- Link to the 'what to expect' content section in top nav
- Remove subtopic from 'what to expect' section (Note: @hursey013 - Can we make it so that we don't have subtopics for this section? Ideally there are no jump downs within this section, but we could always get Amanda to write something else. For now, I have "what to expect" again as a placeholder.)

### DIFF
--- a/src/components/Home.js
+++ b/src/components/Home.js
@@ -67,10 +67,10 @@ class Home extends Component {
               <h3 css={tw`text-4xl font-light text-gray-700`}>
                 What to expect
               </h3>
-              <h4 css={tw`text-xl font-light text-gray-700`}>
+              <span css={tw`block text-xl font-light text-gray-700`}>
                 Help is available, treatment works, and people recover every
                 day.
-              </h4>
+              </span>
             </div>
             <div css={tw`flex flex-wrap -mx-2`}>
               {content.map(card => (
--- a/src/components/Home.js
+++ b/src/components/Home.js
@@ -12,7 +12,7 @@ const content = [
     slug: 'understanding-addiction',
     heading: 'Understanding addiction',
     body:
-      'Addiction is a chronic disease that changes the brain and alters decision-making. Treatment works, help is available, and people recover from addiction every day.'
+      'Addiction is a chronic disease that changes the brain and alters decision-making.'
   },
   {
     slug: 'treatment-options',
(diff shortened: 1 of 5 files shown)


allisonnorman:
Updates for #109

allisonnorman:
**Note:** I was messing around with cards.js to test something earlier and though I undid it and couldn't see it locally, it was pushed in the last commit. I went ahead and changed it again, then undid, and pushed again and it seems to be removed now. Might be worth double checking that.

src/components/Home.js
```diff
@@ -67,6 +67,10 @@ class Home extends Component {
               <h3 css={tw`text-4xl font-light text-gray-700`}>
                 What to expect
               </h3>
+              <h4 css={tw`text-xl font-light text-gray-700`}>
```
hursey013: Semantically this isn't really a heading, so we can keep it visually the same while changing the markup:
```suggestion
              <span css={tw`block text-xl font-light text-gray-700`}>
```

src/components/Home.js
```diff
@@ -67,6 +67,10 @@ class Home extends Component {
...
+                Treatment works, help is available, and people recover every
+                day.
+              </h4>
```
hursey013:
```suggestion
              </span>
```

src/components/Home.js
allisonnorman: Change made

src/components/Home.js
allisonnorman: Change made

allisonnorman:
@hursey013 - Made those changes. Do you think you could do the other two things and we can push this all at once or would you prefer for me to just keep the issue open for those?

- Link to the Getting started content section in top nav
- Remove subtopic from Getting Started section (Note: @hursey013 - Can we make it so that we don't have subtopics for this section? Ideally there are no jump downs within this section, but we could always get Amanda to write something else. For now, I have "what to expect" again as a placeholder.)

allisonnorman:
Looks good to me! Feel free to merge, @hursey013

PR #114, opened 2019-07-18
Set default distance when resetting filters

hursey013:
Fixes https://github.com/18F/samhsa-prototype/issues/112

Include distance in query sent to action when resetting filters, otherwise distance is not included and will return all results rather than the default of 10mi radius.


src/components/Form/Filters.js
```diff
@@ -45,7 +45,12 @@ export class Filters extends Component {
   };
 
   handleReset = () => {
-    this.props.resetAllFilters({ location: this.props.location });
+    const { resetAllFilters, initialValues = {}, location } = this.props;
```
el-mapache: `initialValues` is always guaranteed to exist as at least an empty object, so we don't have to set a default value here

hursey013: I guess I needed this to not break the associated test, so I set it to an empty object in the test instead: https://github.com/18F/samhsa-prototype/pull/114/commits/cd08484486326c92fbc94d8122f7a0b3787a1165 - is there a better way to handle props that are brought in from redux that won't exist in the tests?

el-mapache: That totally makes sense. The wayI would approach this is to test the component as a separate entity from connected component.

In the `Filters` files, add:
```
export { Filters }
```
and import that into your test. Then, you can directly supply the the props the component is expecting to receive from `connect`'s `mapStateToProps` function.

Issue #62, opened 2019-07-09
Add functionality for "reporting bad data" via google analytics


allisonnorman:

Jkrzy:
We'll know we're done when:

- [ ] A click on the "Report a problem with this listing" button generates an `Event` containing the `frid` of the facility
- [ ] After clicking, the user receives a confirmation that the issue has been reported for follow-up and the record for review.
- [ ] A "report" can only be submitted once per page visit, the button should be disabled thereafter.

Jkrzy:
The current implementation of this is a proof-of-concept leveraging google analytics to track identifiers of reported facilities. We'll be evaluating how often this feature is used and if necessary migrating to a more robust implementation down the line.

PR #117, opened 2019-07-18
Add reported facility to store

hursey013:
Fixes https://github.com/18F/samhsa-prototype/issues/62

This adds some visual feedback for when a facility is reported:
Screen Shot 2019-07-18 at 3 56 31 PM
...and limits reporting to once per session. The caveat being since it's not currently stored anywhere, reloading the app would allow someone to report the facility again. Future PR could make use of localStorage to make remembering what facilities have been reported more sticky.

This needs some tests, but will follow up with them in a future PR as well.


Jkrzy:
Functionality fulfills request of #62.

PR #130, opened 2019-07-22
Add loading animation

hursey013:
Fixes https://github.com/18F/samhsa-prototype/issues/82


allisonnorman:
@hursey013 - I noticed that before, the filter section loaded while the user awaited results, but this change removes the filter section while waiting for results. Was this intentional? I think it could actually be helpful to pre-load the filters to indicate what the next step will be and sort of gives a "you're doing the right thing" kind of feel but curious what you think/if that's feasible.

hursey013:
I removed it because in the event that someone is actually seeing this loading screen, and are perhaps on a slow connection, I didnt want to encourage interacting with the filters while the endpoint is still being interacted with.

Brooklyn shared a screen capture of her on the phone a while back where she did a search from the homepage and then the filters were visible while the results were loading on the following page and she thought that the search didnt work and hit the search button again - that's what I was attempting to avoid.

For most people, the results and the filters are going to be presented almost instantly, but happy to tweak this if you think it's beneficial.

allisonnorman:
No, that makes a lot of sense. This is great and we will see if we get more feedback / observe that what I originally suggested needs to be revisited later as we build out the UX.

Issue #101, opened 2019-07-16
Add address validation error


allisonnorman:
Recommended from Ben : if clicked without entering a known location, show an alert with a helptext reminder of the search format required. For example: 

![Screen Shot 2019-07-16 at 1 54 56 PM](https://user-images.githubusercontent.com/6900926/61329085-53f3c980-a7d1-11e9-91f9-3c99d3b6cacb.png)

Jkrzy:
Waiting on content/messaging review in #141

hursey013:
@Jkrzy Ok to close?

PR #141, opened 2019-07-24
Provide instructions/warning if input on homepage doesn't yield any Autocomplete results

Jkrzy:
For #101, ideating around possible implementations.

This approach, absent styling and content review of messaging, renders a note when:

A user has provided input in the `location` field on the homepage
Given that input, the Autocomplete API is unable to provide any location suggestions

Screen Shot 2019-07-24 at 11 09 16 AM


src/components/Input/Location.js
```diff
@@ -5,6 +5,11 @@ import tw from 'tailwind.macro';
 import Label from './Label';
 
 class Location extends Component {
+  constructor(props) {
```
hursey013: Using the binding ability of arrow functions, this constructor can be written as:
```
state = {
  noResultsWarning: false
}
```

hursey013:
I think this is a definitely a great way to leverage the built-in functionality of the library. We do have similar logic occurring in the FormHomepage component itself however, which is how we are currently able to disable the submit button: https://github.com/18F/samhsa-prototype/blob/master/src/components/Form/FormHomepage.js#L46. I'm wondering what you think about leveraging that and displaying the error message there, vs having it tied into the Location input itself?

Jkrzy:
Thanks @hursey013! My thought was, keep the validation as close as possible to the source data and allow us to re-use this everywhere we're leveraging the `Location` component.

Jkrzy:
Took another shot at the styling here @hursey013 , thoughts?

hursey013:
I think adding margin to the button on the homepage could work in some cases but I don't think it's 100% reliable. It may change, but right now we're using a system font stack, so the font will change depending on the OS which means a hardcoded number of pixels will not always line up correctly. This would also add margin to the top of the button on mobile when the inputs are stacked, so we'd need to do a screen size check and conditionally add the margin.

I think I'd still lean towards updating the state in the parent as being the most "reacty" and flexible option since it will allow us to have full control over the placement and styling based on whether its on the homepage or results page. Example implementation for homepage: https://github.com/18F/samhsa-prototype/commit/c36eebf49be5e16961681f014b6309ccf09b3cbb

Jkrzy:
Thanks @hursey013! Your approach takes the good parts here and improves quite a bit. Closing this and re-opening with your branch.

Jkrzy:
Updated per @hursey013 comments above. Also implemented for Results page.

I think we're ready for  content 👀.

@amandacostello @bpdesigns , can you take a glance at the messaging here?

### Desktop - homepage
Screen Shot 2019-08-02 at 9 19 04 AM

src/utils/warnings.js
```diff
@@ -0,0 +1,2 @@
+export const LOCATION_WARNING =
+  'Please enter a city or postal code and select one of the results. If you need help finding treatment, call us 1-800-662-HELP (4357)';
```
hursey013: I forgot that I pull the phone number out into a constant too if you want to use it here. Example of it being used on the 404 page: https://github.com/18F/samhsa-prototype/blob/master/src/components/NoMatch.js#L22-L30

Jkrzy: Great catch, thanks @hursey013 !

bpdesigns:
@amandacostello any thoughts on the wording above? I think it could be shorter and just read 

> Please enter a city or zip code

amandacostello:
> @amandacostello any thoughts on the wording above? I think it could be shorter and just read
> 
> > Please enter a city or zip code

Agree with this; since it returns results in territories, this is more inclusive.

Jkrzy:
Thanks @bpdesigns and @amandacostello!

One last look please @hursey013 and I think we're OK to merge.

PR #143, opened 2019-07-24
Deep link facility details

keithnorm:
This is a workaround due to the API not supporting a GET /listings/:id endpoint.

It works by generating url's in the format of `/details/:slug-:hashed-id/:geo:distance`. I wasn't sure if all listing names were unique so that is why I appended `:hashed-id` onto the slugified name. 

With that info in the url, when the app routes to the details page, it does 2 queries:
1) the first query fetches pageSize=1 with the given lat/lng and distance
2) the second query fetches pageSize=totalRecordCount as returned from the first query, to guarantee that the listing being viewed is in the returned result set

This also makes a change to reload results whenever there is a change in the form filters.


keithnorm:
If this is something y'all might actually consider integrating, I might suggest to pull the geo/distance params out of the url path and put it into query params instead, and maybe default to fetching all records if someone hits a url without query params so that it always works (though would be a performance hit in the latter case).

My thinking was that this page _requires_ those params in order to render, so it has to be part of the path. But, it's breaking a best practice to have multiple url's for the same page, and if the API ever does support an endpoint to return a single location, then supporting the base url path of /location/:slug-:id could still be supported, whereas supporting url's with extra params in the path would be more onerous.

Jkrzy:
Thanks @keithnorm! Greatly appreciate the contribution here and hope that you'll follow along w/ our progress. 

We're going to pull this branch into the repo and open a new PR, allowing [Federalist](https://federalist.18f.gov/) to build us a preview. Will update here w/ reference and close this PR when that happens.

keithnorm:
Sounds good, thank you @Jkrzy! I will indeed be following along :)

Issue #126, opened 2019-07-19
Implement SU/MH-specific filters


allisonnorman:
Background: Mental health and substance use are closely tied together. SAMHSA's treatment locator currently returns both facilities that treat both SU and MH, and they are classified as MH or SU based on which survey they have filled out. 

While our focus during the engagement has been on substance use, we want to understand how people are engaging with these filters in order to build a longer term strategy (to either bring in MH facilities, to create a separate locator, etc).  This testing approach has trade offs, as the content and branding of the site is on substance use treatment. For this reason, we have decided to only include co-occurring (SU+MH) and a broad MH services filter to get a signal on whether people are looking for mental health support through this tool.  

User stories:
1. As a visitor looking for substance use treatment, I should not be shown mental health facilities returned by default unless I select a mental-health specific type of care filter (mental health or co-occurring) so that I do not reach out to a facility that might not specialize in providing substance use treatment disorder. 
2. As a visitor looking for co-occurring treatment, I should be able to select co-occurring under type of care and see both mental health and substance use facilities so that I can connect with a treatment providers who can support my MH+SU-specific needs.
3. As a visitor looking for mental health services, I should be able to find mental health facilities under type of care. 
4. As a visitor looking for psychiatric emergency walk in services, I should be able to find mental health facilities under type of care 
5. As a visitor not sure of what I'm looking for, I should be able to read about co-occurring and mental health services types of care in the informational content - @amandacostello, Kait

Summary (alternately, see notes [here](https://docs.google.com/document/d/14S88KXIqK4VeLheoqmRgXa-OiHo3rpFGRqW3u3Fs7xk/edit#heading=h.5cj4phf568f8) 
1. Remove mental health facilities in default search. Unless a visitor interacts with one of the below filters, they should not see mental health facilities returned. 
2. Adjust how results are returned for following filters in the 'type of care' dropdown:
- 'Co-occurring mental health and substance use treatment' (service codes, CO, SA, MH)  Return either SU or MH results with the service code CO
- 'Mental health services only' (service code - MH): Return all MH services only (no SU)
- Psychiatric emergency walk-in services (service code - PSY): This filter is MH-only. If SU is removed from default search, this will need to be set up to return MH results via the filter query.

hursey013:
For posterity, 'Co-occurring mental health and substance use treatment' would be `sType: BOTH, sCodes: CO`

allisonnorman:
[...]

PR #149, opened 2019-07-25
Support MH filters, refactor API param builder

hursey013:
Fixes https://github.com/18F/samhsa-prototype/issues/126

This adds the new filters for mental health.

Also refactors how queries are prepared to be sent to the endpoint in order to make it easier to understand and will help support future filter changes.


src/utils/api.js
```diff
@@ -9,66 +9,47 @@ export default axios.create({
...
+  const codeValues = Object.values(codes);
+
+  const params = {
+    sType: parseType(type),
+    sCodes: parseCodes(type, codeValues),
+    pageSize: DEFAULT_PAGE_SIZE,
+    page: page || 1,
+    sAddr: location && `${location.location.lat}, ${location.location.lng}`,
+    limitType: distance && 2,
+    limitValue: distance && distance,
```
el-mapache: Here, I think it's clearer to put in `distance || null`. The resulting value is the same, and its consistent with what you've done on line 20.

src/utils/api.js
```diff
@@ -9,66 +9,47 @@ export default axios.create({
...
+    sLanguages: language && language
```
el-mapache: The above comment applies here as well

src/utils/api.js
```diff
@@ -9,66 +9,47 @@ export default axios.create({
...
+  const { distance, language, location, page, type, ...codes } = query;
+  const codeValues = Object.values(codes);
+
+  const params = {
+    sType: parseType(type),
+    sCodes: parseCodes(type, codeValues),
+    pageSize: DEFAULT_PAGE_SIZE,
+    page: page || 1,
+    sAddr: location && `${location.location.lat}, ${location.location.lng}`,
+    limitType: distance && 2,
```
el-mapache: I think it would be helpful to make the `limitType` value of `2` a named constant. This can help document what the code is supposed to be doing to someone unfamiliar with the codebase.

src/utils/api.js
```diff
@@ -9,66 +9,47 @@ export default axios.create({
...
+        .concat(type)
+        .filter(code => !!code)
+        .toString()
+    : codes.toString();
+};
 
-  if (!params.sCodes) {
-    const { sCodes, ...finalParams } = params;
+const parseType = type => {
+  if (type === 'MH' || type === 'WI') {
```
el-mapache: I think it would be helpful to have each of these types in an enum, with descriptive names. Reading this code, I don't know what type refers to, and I don't know what each of these values are supposed to represent.

Something like:

```
const descriptiveTypeExplanation = {
  descriptiveTypeName1: 'MH',
  ...etc
}
```

Then, in this function, you can call e.g.:
`if (descriptiveTypeExplanation.descriptiveTypeName1) { return 'my value' }`

I think the function could be name a bit more verbosely to explain exactly what the 'types' are, or do.

src/utils/api.js
```diff
@@ -9,66 +9,47 @@ export default axios.create({
...
-        page: value
-      };
-    }
+  return params;
+};
 
-    if (key === 'type' && value === 'Intake') {
-      return memo;
-    }
+const parseCodes = (type, codes) => {
```
el-mapache: Here, I think a more descriptive name and potentially a comment to explain what the 'codes' are that are being parsed, and what they control, would be super helpful for future devs (or even you!) on this project.

src/utils/api.js
```diff
@@ -9,66 +9,47 @@ export default axios.create({
...
-    return {
-      ...memo,
-      sCodes: memo.sCodes
-        .split(',')
-        .filter(v => !!v)
-        .concat(value)
-        .join()
-    };
-  }, initialValues);
+  return !skipTypeCodes.includes(type)
```
el-mapache: This is personal preference, but I think the inverse of this statement is easier to read

`skipTypeCodes.includes(type) ? codes.toString : filterchain`

src/utils/api.js
```diff
@@ -9,66 +9,47 @@ export default axios.create({
...
-      };
-    }
+  return params;
+};
 
-    if (key === 'type' && value === 'Intake') {
-      return memo;
-    }
+const parseCodes = (type, codes) => {
+  const skipTypeCodes = ['Intake', 'MH'];
```
el-mapache: This constant can live outside the function, no reason to reallocate the array every time `parseCodes` is called.

src/utils/api.js
```diff
@@ -9,66 +9,47 @@ export default axios.create({
...
+    pageSize: DEFAULT_PAGE_SIZE,
+    page: page || 1,
+    sAddr: location && `${location.location.lat}, ${location.location.lng}`,
+    limitType: distance && 2,
+    limitValue: distance && distance,
+    sLanguages: language && language
+  };
+
+  Object.keys(params).forEach(
+    key => (params[key] == null || !params[key]) && delete params[key]
```
el-mapache: This statement will omit params that have a value of `0`, is that the intent? If so, the check for `!params[key]` will also cover a `null` value, and the first check is superfluous.

If both checks are necessary, there should be a `===` in the first half of the predicate.

PR #150, opened 2019-07-26
Clean up homepage

hursey013:
Pulls heading and blurb for the homepage cards from the existing content.js file since it the content is already in there, eliminating the need to duplicate it in the Home component

Other small consistency/styling improvements.


hursey013:
@allisonnorman yes, this would consolidate the content into one place, and share it between the homepage cards and the info pages. The homepage card does required the `description` value in `content.js` (which is also the lead paragraph text on the info page), but if that doesn't work let me know and we can find a workaround.

Issue #133, opened 2019-07-22
Update home page design to incorporate 4th content box for mental health


allisonnorman:
Based on SAMHSA leadership feedback, we need to incorporate aspects of mental health in our prototype to test with the public. This will include updates to info content, which we will want to link to from the homepage.

What we'll need from a design perspective from @nickttng: 
- A new icon for mental health box (we can use lorem ipsum for the content for now)
- Update sizing of all 4 boxes / update proportions, if needed

hursey013:
I would love to see an alternative to adding a fourth box - to me, it feels veery cramped with four boxes, so perhaps there is a better way to display that information? The boxes are really just an artifact from rapid prototyping, so I don't think it's a pattern we need to follow if there are better options.

nickttng:
@allisonnorman @hursey013 The mental health icon (SVG + PNGs labeled _what-to-expect-mental-health_ is now in the shared Drive: `9. Assets > Visual/UI > Icons > What to expect`

In addition, the new _what-to-expect-addiction.svg_ and PNGs (tangled brain) to swap out the old (cycle arrow)

Background context/convo in 🔒[Slack](https://gsa-tts.slack.com/archives/CGBQQ8KFD/p1564169848038300)

hursey013:
Closed by https://github.com/18F/samhsa-prototype/pull/156

Issue #151, opened 2019-07-26
Make changes to support MH content


allisonnorman:
- [X] Add MH info content
- [X] Update lede on home page (same as subtitle on MH page)
- [X] Add new MH filter descriptions to "treatment options" section for MH services
- [X] Add new MH filter descriptions to "treatment options" section for co-occurring
- [X] Update "Quickly find providers who treat substance use disorders and addiction" on homepage to reflect MH

PR #156, opened 2019-07-26
Add mental health content

allisonnorman:
Making some changes for #151 
- Created new MH section on info content page + subtitle for home page
- Updated 'treatment options' with descriptors for new filters
- Updated subtitle under Find treatment to include mental illnesses


allisonnorman:
@hursey013 - The MH content is all in. Amanda is going to do some overall side by side comparison of the content to make sure it's all in place but I think we're good to merge this in.

I do realize it will affect the formatting of the home page as the fourth box is getting added. Wondering if it makes sense to shrink those boxes down and add the icon for now while we wait on the final design from Nick, or if it's better to use this preview link until the design is ready. Up to you.

src/utils/content.js
```diff
@@ -230,6 +230,145 @@ export default () => [
...
+                <li>Hyperactive behavior</li>
+                <li>Frequent nightmares</li>
+                <li>Frequent disobedience or aggression</li>
+                <li>Frequent temper tantrums</li>
+              </ul>
+            </p>
+            <p>
+              For many people, getting an accurate diagnosis is the first step
+              in a treatment plan.{' '}
+              <a href="https://www.nimh.nih.gov/health/topics/integrated-care/index.shtml">
```
amandacostello: We can delete this link! It was left in as a reference in the working doc.

hursey013:
Added new icon. Fixes https://github.com/18F/samhsa-prototype/issues/133

Issue #158, opened 2019-07-29
Change default location search radius


bpdesigns:
Current state:  The location search defaults to a 10mile radius

Desired state: Update the location search default to a 25 mile radius.

Rationale: Rural areas may not have results within 10 miles. When that happens it may confuse or discourage people to search

bpdesigns:
@allisonnorman @hursey013

PR #159, opened 2019-07-29
Change default search radius (25mi)

hursey013:
Fixes https://github.com/18F/samhsa-prototype/issues/158

Previously 10 miles, now 25 miles from center point of location.


Jkrzy:
Tweaked this slightly to self-document the conversion from miles to meters. What do you think @hursey013?

hursey013:
Good call!

Issue #163, opened 2019-07-30
Results are being returned within 25 miles, but the filter displays 5 miles


allisonnorman:
When I was looking at the changes made in #159, I noticed that results are being returned within 25 miles, but the filter displays a distance radius of 5 miles. I'm not sure if this is by design but I think we should reflect the radius we are applying in the distance drop down. 

![Screen Shot 2019-07-30 at 9 13 37 AM](https://user-images.githubusercontent.com/6900926/62146613-9a095c80-b2aa-11e9-9e38-21af4581ca14.png)

PR #165, opened 2019-07-30
Fix for distance dropdown

hursey013:
Fixes https://github.com/18F/samhsa-prototype/issues/163

### DIFF
--- a/src/components/Form/FormFilters.js
+++ b/src/components/Form/FormFilters.js
@@ -24,7 +24,7 @@ const Row = styled.div`
 `;
 
 const RowFlex = styled.div`
-  ${tw`flex flex-wrap -mx-3 mb-6`}
+  ${tw`flex flex-wrap -mx-2 mb-6`}
 `;
 
 const Form = styled.form`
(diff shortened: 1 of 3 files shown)


hursey013:
@allisonnorman I'd be hesitant to make the location input box much smaller, do you think it getting cut off warrants moving the dropdown down onto it's own line? Would abbreviating miles to "mi" be ok?

allisonnorman:
I think it's fine if miles gets cut off to mi, but mi is getting cut off too. Would it be weird to put miles in the label for the drop down? "Distance (in miles)" and remove miles from the drop down?

allisonnorman:
Or, is it possible to show miles only when the user is looking at the drop down options, but the selected option doesn't show miles? (Maybe I'm overthinking about this but throwing out options. We could always hold off to see if Nick or Austin could think about this more?)

Jkrzy:
What about moving the units to the form label?
Screen Shot 2019-07-30 at 3 42 44 PM

hursey013:
@allisonnorman increased size of dropdown to make sure all values are visible.

PR #167, opened 2019-07-30
Add function to convert services array into object

hursey013:
This converts the services data returned with each facility from an array like:
```
services:[
{f1: "Type of Care", f2: "TC", f3: "Substance use treatment"},
{f1: "Service Settings (e.g., Outpatient, Residential, Inpatient etc.)", f2: "SET", f3: "Outpatient; Intensive outpatient treatment"},
{f1: "Type of Opioid Treatment", f2: "OT", f3: "Do not use medication for opioid addiction"}
...
```
into an object accessible by service code:
```
{
AGE: {name: "Age Groups Accepted", values: ["Adults", "Young adults"]},
AS: {name: "Ancillary Services", values: ["Case management", "Consumer-run (peer-support) services", "Self-help groups", "Social skills development"]},
ASSES: {name: "Assessment/Pre-treatment", values: ["Comprehensive substance use assessment", "Outreach to persons in the community"]}
...
```
So rather than having to guessing the index of the value in an array like `services[8].f3`, they can be accessed by service code: `services.AGE.values`. This eliminates the chance of missing services changing the index number and is more readable overall.

This PR shows a proof of concept of this method by displaying the `Type of Care` value on the results facility card.

### DIFF
--- a/src/components/Details.js
+++ b/src/components/Details.js
@@ -20,21 +20,16 @@ export class Details extends Component {
     this.props.history.goBack();
   };
 
-  renderService = ({ name, values }) => {
-    console.log(name, values);
-    return (
-      <div css={tw`mb-4 pb-4 border-b`} key={name}>
-        <h3 css={tw`font-semibold text-sm`}>{name}</h3>
-        <ul
-          css={tw`text-sm leading-relaxed text-gray-700 list-disc list-inside`}
-        >
-          {values.map((item, index) => (
-            <li key={index}>{item}</li>
-          ))}
-        </ul>
-      </div>
-    );
-  };
+  renderService = ({ name, values }) => (
+    <div css={tw`mb-4 pb-4 border-b`} key={name}>
+      <h3 css={tw`font-semibold text-sm`}>{name}</h3>
+      <ul css={tw`text-sm leading-relaxed text-gray-700 list-disc list-inside`}>
+        {values.map((item, index) => (
+          <li key={index}>{item}</li>
+        ))}
+      </ul>
+    </div>
+  );
 
   reportFacility = () => {
     const { facility } = this.props;
--- a/src/components/Details.js
+++ b/src/components/Details.js
@@ -20,16 +20,21 @@ export class Details extends Component {
     this.props.history.goBack();
   };
 
-  renderService = service => (
-    <div css={tw`mb-4 pb-4 border-b`} key={service.f2}>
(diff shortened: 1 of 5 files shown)


Jkrzy:
This is definitely an improvement over accessing the `serviceCategories` by index.  This approach preserves the `category: servicesArray` structure which doesn't allow us to access `services` directly. 

That's OK as I don't believe we've seen an explicit need for that ability but I don't think its reflected in the original comment above. 

One thing to flag:

What's happening here is unclear to anyone not familiar with the coded values of the `serviceCategories`. Can we make the intent clear?
 ```diff
-<p>
-  <span css={tw`font-semibold`}>Type of care:</span> {services[0].f3}
-</p>
+<p>{renderService(services.TC)}</p>
```

**Comment by el-mapache** [comment]
Echoing @Jkrzy's comment, this is definitely an improvement over accessing this data via array. 

If we plan on using the service codes in more than one place (and I think we do?), I'd probably just transform the data once it's fetched from the server, rather than generating it for each component that needs it.

I also agree with the last point in Joe's comment. Something like `services.typesOfCare` is more explicit, and we can store a reference to the actual service code in the accessor object.

**Comment by hursey013** [comment]
@Jkrzy yeah you are right that this mainly improves being able to access service by category (Age Groups Accepted, Ancillary Services, Assessment/Pre-treatment, etc) - the values within each of those categories are just semicolon delimitated strings, so I'm converting those into arrays where applicable, but I'm not sure where I'd go from there unless we wanted to have some sort of lookup table with all know values of services that we could then map to key values. Do you have any thoughts on how we could get any more granular?

**Comment by Jkrzy** [comment]
Let's move this ongoing discussion of how to best retrieve, store, and access service information to a new issue.

**Comment by hursey013** [comment]
@el-mapache moved the transform to the reducer so that it's accessible to all components

File: src/components/Details.js [hunk 1]
```diff
@@ -20,16 +20,21 @@ export class Details extends Component {
...
-      <h3 css={tw`font-semibold text-sm`}>{service.f1}</h3>
-      <ul css={tw`text-sm leading-relaxed text-gray-700 list-disc list-inside`}>
-        {service.f3.split('; ').map((item, index) => (
-          <li key={index}>{item}</li>
-        ))}
-      </ul>
-    </div>
-  );
+  renderService = ({ name, values }) => {
+    console.log(name, values);
```
el-mapache: stray console log here!

PR #168, opened 2019-07-30
Updated spacing

amandacostello:
Fixed spacing issues around <strong> and <a> tags where words were running together.


Jkrzy:
Thanks @amandacostello !!!

allisonnorman:
Looks like some of the spacing didn't work. I have had issues with this too and it's one of those things that doesn't necessarily make sense just looking at the code so I will make a couple updates locally too to verify and push them up. Thanks to @amandacostello's diff, it'll be easy to do but it's something we should be careful of checking in the preview link since it's not super straightforward.

allisonnorman:
Oh wait, nevermind - these all look great, I just saw a couple other ones! :)

Issue #43, opened 2019-07-02
Add error page


hursey013:
In the event that the vendor's endpoint is down or there are other technical issues, we should provide messaging to the user instead of the default "no results" message we currently have.

hursey013:
@allisonnorman @amandacostello Do you have any suggestions for content for this page? Basically someone would arrive here if they initiated a search but for some reason something goes wrong getting the data back (the vendor's servers are down for example).

allisonnorman:
What about using the 404 error language that we have today? (It's a slightly different version of the no results messaging). Thoughts @amandacostello?

amandacostello:
Idea for content:

**Something's wrong on our side.**

We're having technical issues pulling up search results right now. The problem should be fixed soon, and in the meantime you can call XXX-XXXX (Helpline) or XXX-XXXX (Crisis Line) for immediate help.

amandacostello:
THOUGH, even if they call the help line, can they still get help if the servers are down?

Jkrzy:
@amandacostello great question! We've got confirmation from SAMHSA that the helpline is not impacted by availability of the existing TL. We can and should direct folks to call if we're unable to retrieve data.

PR #172, opened 2019-08-01
Add error page for when endpoint is down

hursey013:
Fixes https://github.com/18F/samhsa-prototype/issues/43

Adds error page in case vendor endpoint is down. Including screenshot of page since there is no easy way to reproduce on preview site:

Screen Shot 2019-08-01 at 1 39 00 PM

### DIFF
--- a/src/components/Error.js
+++ b/src/components/Error.js
@@ -1,0 +1,34 @@
+import React from 'react';
+import tw from 'tailwind.macro';
+import 'styled-components/macro';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faPhone } from '@fortawesome/free-solid-svg-icons';
+import { OutboundLink } from 'react-ga';
+
+import { HELPLINE_LINK, HELPLINE_TEXT } from '../utils/constants';
+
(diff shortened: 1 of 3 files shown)


src/components/Error.js
```diff
@@ -0,0 +1,34 @@
...
+  <div css={tw`mx-auto p-6 mb-6 text-center`}>
+    <h2 css={tw`text-3xl font-bold mb-10`}>Something's wrong on our side.</h2>
+    <div css={tw`max-w-xl mx-auto`}>
+      <p css={tw`text-xl mb-6`}>
+        We're having technical issues right now. The problem should be fixed
+        soon, and in the meantime SAMHSA's national helpline is available 24/7
+        and can assist you with treatment referrals and information.
+      </p>
+      <OutboundLink
+        eventLabel="Helpline link from 404"
```
Jkrzy: Lets differentiate this from the 404 tracking event

Jkrzy:
Would be great to track this in GA, sending an event when rendered.

hursey013:
Oops, added the right label for the GA event @Jkrzy

Issue #173, opened 2019-08-01
Update results automatically on any filter change 


allisonnorman:
Today, users need to click the "Update providers" button when they change the filters. Based on user feedback and some of the nuances of the advanced filters getting cleared when the section is collapsed, we'd like to have results update automatically as users engage with the filters.

User story:
As a user, I want to be able to see the results update as I engage with filters without having to select a button so that I do not forget or misunderstand how to update the results.

PR #174, opened 2019-08-01
Search facilities immediately after filter change

hursey013:
Fixes https://github.com/18F/samhsa-prototype/issues/173


hursey013:
Updated to preserve the ability to submit the filters with a button on mobile.

I also updated the interaction with the location input to not empty on focus to avoid accidental clearing. This was the pattern I observed on BCBS and seems more appropriate when updating the results automatically.

PR #180, opened 2019-08-05
[WIP] Deep link facility details

Jkrzy:
Bringing PR https://github.com/18F/samhsa-prototype/pull/143 and associated branch in-house.

> @keithnorn said:
> This is a workaround due to the API not supporting a GET /listings/:id endpoint.
> 
> It works by generating url's in the format of /details/:slug-:hashed-id/:geo:distance. I wasn't sure if all listing names were unique so that is why I appended :hashed-id onto the slugified name.
> 
> With that info in the url, when the app routes to the details page, it does 2 queries:
> 
> the first query fetches pageSize=1 with the given lat/lng and distance
> the second query fetches pageSize=totalRecordCount as returned from the first query, to guarantee that the listing being viewed is in the returned result set
> This also makes a change to reload results whenever there is a change in the form filters.


Jkrzy:
Replaced by #215.

Issue #187, opened 2019-08-05
Text should have a color contrast ratio of 4.5:1 or greater with the background.


Jkrzy:
A few text color/contrast issues: https://accessibility.18f.gov/color/

Tested w/ https://developer.paciellogroup.com/resources/contrastanalyser/

Grey text on white background is 4:1 contrast ratio.
Results page information is 1.5:1
<img width="418" alt="Screen Shot 2019-08-05 at 4 45 32 PM" src="https://user-images.githubusercontent.com/3485564/62494094-aa38a480-b7a0-11e9-9c4a-7c40d179f172.png">
<img width="523" alt="Screen Shot 2019-08-05 at 4 45 36 PM" src="https://user-images.githubusercontent.com/3485564/62494127-be7ca180-b7a0-11e9-9746-d811a98c2fc8.png">
<img width="344" alt="Screen Shot 2019-08-05 at 4 48 36 PM" src="https://user-images.githubusercontent.com/3485564/62494209-e8ce5f00-b7a0-11e9-92bb-317017dbe546.png">

_Originally posted by @Jkrzy in https://github.com/18F/samhsa-prototype/issues/184#issuecomment-518394273_

Contrast ratio: 4.03:1
![Screen Shot 2019-08-06 at 10 58 39 AM](https://user-images.githubusercontent.com/3485564/62551021-3b5b5a00-b839-11e9-82d0-019980f73882.png)

Jkrzy:
Closing this w/ the merge of #193.

We're expecting additional design changes and will revisit this if necessary.

PR #193, opened 2019-08-06
Improve contrast by darkening text and backgrounds

Jkrzy:
Herein, for #187, we darken the grey used as `text-color` and blue used as button backgrounds to reach a desired contrast ratio of 4.5:1 throughout the site.

### DIFF
--- a/src/components/Input/Button.js
+++ b/src/components/Input/Button.js
@@ -10,14 +10,14 @@ const Button = styled.button`
 
   ${props => props.secondary && tw`w-full bg-gray-100 hover:bg-gray-200 border`}
 
-  ${props => props.outline && tw`text-blue-700 border border-blue-500`}
+  ${props => props.outline && tw`text-blue-700 border border-blue-700`}
 
   ${props =>
     props.base &&
     tw`hover:text-gray-900 text-gray-900 bg-gray-300 hover:bg-gray-400`}
 
   ${props =>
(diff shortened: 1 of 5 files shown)


src/components/Input/Button.js
```diff
@@ -6,11 +6,11 @@ const Button = styled.button`
 
   ${props =>
     props.primary &&
-    tw`bg-blue-500 hover:bg-blue-700 border border-blue-500 text-white`}
+    tw`bg-blue-700 hover:bg-blue-900 border border-blue-700 text-white`}
 
   ${props => props.secondary && tw`w-full bg-gray-100 hover:bg-gray-200 border`}
 
-  ${props => props.outline && tw`text-blue-600 border border-blue-500`}
+  ${props => props.outline && tw`text-blue-700 border border-blue-500`}
```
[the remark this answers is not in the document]
hursey013: We also need to change the `disable` state of the button to: `hover:bg-blue-700` so it matches the `primary` button.

src/components/Pagination.js
```diff
@@ -60,6 +60,7 @@ class Pagination extends Component {
           pageRangeDisplayed={3}
           onPageChange={this.handlePageClick}
           forcePage={page - 1}
+          css={tw`text-purple-800`}
```
hursey013: I don't think this will get applied to anything - what should inherit this?

Jkrzy: whoops, that shouldn't be in there 🤦‍♂. Good catch, thanks!

Issue #189, opened 2019-08-06
Users should be able to tab once to select and move-past button links.


Jkrzy:
Users should be able to tab once to select and move-past button links.

Most of the <button> elements on the site are wrapped by <Link> components. This renders as `<a><button></button></a>` which requires tabbing through both the anchor and the button element.

_Originally posted by @Jkrzy in https://github.com/18F/samhsa-prototype/issues/184#issuecomment-518676440_

PR #205, opened 2019-08-07
Single focusable element for facility details links

Jkrzy:
For #189, we render the links to facility details from the Results page as a single anchor element. 

This includes a small visual change as we're falling back on the existing styling for ` <Button 
as={Link} outline="true" ...`

Screen Shot 2019-08-06 at 10 58 39 AM

### DIFF
--- a/src/components/Card.js
+++ b/src/components/Card.js
@@ -111,7 +111,7 @@ const Card = props => {
         as={Link}
         to={{ pathname: '/details', state: { frid } }}
         css={tw`print:hidden`}
-        outline="true"
+        primary="true"
       >
         View provider details
       </Button>
--- a/src/components/Card.js
+++ b/src/components/Card.js
@@ -107,12 +107,14 @@ const Card = props => {
         zip={zip}
       />
       <CardDetails phone={phone} website={website} services={services} />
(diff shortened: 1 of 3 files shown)


src/components/Card.js
```diff
@@ -107,12 +107,14 @@ const Card = props => {
         zip={zip}
       />
       <CardDetails phone={phone} website={website} services={services} />
-      <Link
+      <Button
+        as={Link}
         to={{ pathname: '/details', state: { frid } }}
         css={tw`print:hidden`}
+        outline="true"
```
hursey013: Good catch! I think if we want to keep the button style the same we could pass `primary="true"` here rather than `outline`.

Jkrzy: Ah right, lets go with that and reduce the scope of the changes here.

Jkrzy: Will require some styling tweaks

hursey013: I think adding `hover:text-white` to the primary button styling will give us the desired effect.

Jkrzy: Thanks, made those changes and rebased against `master`.

Issue #146, opened 2019-07-25
Users should be able to navigate to a facility page directly 


allisonnorman:
As a visitor looking for treatment for myself or someone else, I want to be able to navigate to a facility details page directly so that I can: 
- Come back the specific facility page at a later time
- Open the details page in another tab/window
- Send the facility page link to someone else 

There are limitations to the back end as there is no endpoint to call to return all facility details. We are exploring workarounds. These options will work for a proof of concept, but will involve the URL changing as facility location details get updated. As such, we should document the workaround we use and share the ask with ET for a long-term, sustainable solution.

hursey013:
Moving this back into the backlog column for the moment until we hear back from vendor on alternative approach.

Jkrzy:
We'll move forward on this with a workaround implementation. 

cc: @hursey013

PR #215, opened 2019-08-08
Route to facility page based on FRID

hursey013:
Fixes https://github.com/18F/samhsa-prototype/issues/146

Piggybacking off of @keithnorm's workaround here: https://github.com/18F/samhsa-prototype/pull/180 this adds the ability to navigate directly to a facility page, enabling folks to share/bookmark specific facilities.

By using a path of `/details/:frid/` along with the lat and long appended as query parameters, we are able to use the existing endpoint to search by lat/long (limiting results to only facilities with the exact lat/long), and then further filter by FRID in order to narrow the results to a single facility. This is accomplished with a single call to the endpoint.

In the future, if a new FRID specific endpoint is enabled, the `/details/:frid/` path can continued to be used (preserving existing links), and the additional query params can be dropped off.

The main caveat of this approach is that the links are dependent on the lat/long. If a facility's lat/long changes, the previous URL of the facility will no longer function.

### DIFF
--- a/src/actions/facility.js
+++ b/src/actions/facility.js
@@ -1,0 +1,68 @@
+import qs from 'qs';
+
+import API from '../utils/api';
+import { DEFAULT_STYPE } from '../utils/constants';
+
+export const RECEIVE_FACILITY_BEGIN = 'RECEIVE_FACILITY_BEGIN';
+export const RECEIVE_FACILITY_SUCCESS = 'RECEIVE_FACILITY_SUCCESS';
+export const RECEIVE_FACILITY_FAILURE = 'RECEIVE_FACILITY_FAILURE';
+export const DESTROY_FACILITY = 'DESTROY_FACILITY';
+
+export const receiveFacilityBegin = () => {
+  return {
+    type: RECEIVE_FACILITY_BEGIN
+  };
+};
+
+export const receiveFacilitySucess = (data, frid) => {
+  return {
+    type: RECEIVE_FACILITY_SUCCESS,
+    payload: { data },
+    frid
+  };
+};
+
+export const receiveFacilityFailure = error => {
+  return {
+    type: RECEIVE_FACILITY_FAILURE
+  };
+};
+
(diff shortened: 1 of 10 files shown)


hursey013:
Unfortunately because of the way Federalist works, this isn't testable on the preview branch. It will need to be tested locally, or once merged into master.

Jkrzy:
@hursey013, Can you expand on why this isn't testable on the federalist preview?

hursey013:
@Jkrzy Federalist relies on the S3 404 workaround to route pages. You can only define one file to point to for the 404, and we are pointing to `/index.html`, so when you get a 404 on a preview branch, it loads in the master branch's index.html, which references `/index.html`, but since we're on a preview branch, `https://cg-efcd7c90-b365-4336-8b87-6174f1975c99.app.cloud.gov/index.html` does not exist.

allisonnorman:
I don't have docker set up locally to see the results. Maybe we could screen share, @hursey013 or I can preview with @keithnorm.

hursey013:
@allisonnorman added FRID to popup links on google maps

Jkrzy:
Nice work @hursey013!

I see a few opportunities for tweaks as we move from passing data via the Results view. 

We should aim to only request facility data once. If a user is coming to the details page from results, we already have that facility data in the store and should leverage it instead of hitting the backend again.

Also, we shouldn't display the `Back to search results` link if someone has navigated directly to a facility page.

hursey013:
Back to you @Jkrzy. Yeah I debated whether it was worth the added complexity to support the two different ways that data could be accessed, but I think I found an approach that isn't too overly complex and can be removed easily in the future.

The `Details` component will check for the existence of `facilities` data in the store (which will only be present when coming from a results page), and if it exists use that rather than firing off an additional request.

Jkrzy:
Looks great!

There's an easy opportunity to capture some metrics on usage of direct-to-facility links/access via google events. Lets explore that in another issue.

Issue #226, opened 2019-08-13
Error on empty location input


hursey013:
Initial a search from the homepage. On the results page, click into the location input field and backspace until there until the value is completely empty. Observe error message.

This happens because an empty location (`sAddr`) string is being submitted to the production endpoint which causes a 500 error.

PR #227, opened 2019-08-13
Prevent empty location value in filter input

hursey013:
Sending an empty location causes a server error on the production endpoint. This can be seen in the steps outlined in https://github.com/18F/samhsa-prototype/issues/226. This prevents the filters form from submitting without a valid location.

I think there are definitely some bigger UX issues/questions associated with default/empty location behavior though that will need to be explored/tested more thoroughly.

### DIFF
--- a/src/components/Form/FormFilters.js
+++ b/src/components/Form/FormFilters.js
@@ -337,13 +337,13 @@ export default connect(
     enableReinitialize: true,
     destroyOnUnmount: false,
     onChange: (values, dispatch, props, previousValues) => {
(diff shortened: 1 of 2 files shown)


hursey013:
Also, a quick note on testing. For one reason or another, the development endpoint _does_ allow an empty location string, so fully testing this will require merging to `master` so that the production endpoint is used 🙁

hursey013:
Need to make some tweaks to this, moving to `WIP`.

hursey013:
Ok, ready for review 👍

PR #240, opened 2019-08-15
Update 'type of care' filter label to 'type of treatment'

### LINKED ISSUE #238: Update filter for "types of care"
The secondary filter **"Type of Care"** should be renamed **"Types of treatment"**, with a "Not sure?" inline link next to it that goes to the types of treatment section: https://findtreatmentbeta.18f.gov/content/treatment-options#types-of-treatment

allisonnorman:
Fixes #238 

Per @hursey013, we will also need to make changes to the label on the card and details pages (and maybe elsewhere) but I will leave it to him to do that as part of this PR or otherwise (if otherwise, let me know and I can create a separate issue)

### DIFF
--- a/src/components/Form/FormFilters.js
+++ b/src/components/Form/FormFilters.js
@@ -172,8 +172,8 @@ export class FormFilters extends Component {
             <Row>
               <Field
                 name="type"
-                label="Type of care"
-                plural="types of care"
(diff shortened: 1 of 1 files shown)


allisonnorman:
Created #242. I'll go ahead and merge this. Thanks!

Issue #234, opened 2019-08-15
Location select box is not accessible


hursey013:
It's close, it's keyboard accessible, but it is not a good experience for voice software since it does not announce any of the location suggestions. We will need to look into improving the accessibility of the current component or perhaps exploring alternative options.

This library mentioned its WAI-ARIA compliant:https://github.com/hibiken/react-places-autocomplete

PR #241, opened 2019-08-15
Implement react-places-autocomplete

hursey013:
Fixes https://github.com/18F/samhsa-prototype/issues/234

This replaces the https://github.com/ubilabs/react-geosuggest library with https://github.com/hibiken/react-places-autocomplete which is WAI-ARIA compliant and includes the proper `aria` and `role` attributes that help support voice software.

The new component is also not perfect, there are a number of Axe errors appearing in relation to it, but I think it's a more solid foundation to start with and the warnings that are appearing should be fixable. 

You can also test out the existing prototype's location picker using voiceover as compared to this one and I think the benefits will be more obvious.


Jkrzy:
Merging this in advance of remediating the stalled `snyk` check. No vulnerabilities reported by `snyk` at this time for the added package.

PR #245, opened 2019-08-21
Update .env.production

Jkrzy:
Moving to a new GA property for initial release of MVP. Updating production GA UID here to effect that change.


hursey013:
Do we want to wait until we're live to merge this?

PR #246, opened 2019-08-21
Submit location on pressing enter

hursey013:
Allow location input to be submitted by pressing enter.

Addresses feedback:
> Weird thing where typing a zip and hitting enter won’t search - I have to either tab over or click the search button


hursey013:
Oops good catch. Added additional check that it's a valid location before allowing submission.

PR #269, opened 2019-08-22
Add visual styles

hursey013:
Implements visual styles based on @austinhernandez's mockups.


src/utils/filters.js
```diff
@@ -1,59 +1,155 @@
...
-
-export const gender = [
-  { value: 'MALE', label: 'Male' },
-  { value: 'FEM', label: 'Female' }
+  { value: 8046.72, label: '5 miles' },
+  { value: 16093.4, label: '10 miles' },
+  { value: 40233.6, label: '25 miles' },
+  { value: 80467.2, label: '50 miles' },
+  { value: 160934, label: '100 miles' },
+  { value: '', label: '100+ miles' }
```
Jkrzy: Good opportunity to self-document this by doing the conversion from miles to meters here.

src/utils/api.js
```diff
@@ -47,3 +47,14 @@ const combineServiceTypeAndServiceCodes = (type = '', serviceCodes = []) => {
         .filter(code => !!code)
         .toString();
 };
+
+const flatternServiceCodes = sCodes => {
```
Jkrzy: Small type in function name, should be `flattenServiceCodes`

src/components/ResultsList.js
```diff
@@ -33,45 +32,14 @@ export class ResultsList extends Component {
...
-            <span css={tw`text-lg text-gray-700 font-light`}>
-              Treatment providers near you
-            </span>
-          </h1>
-          <span css={tw`block text-gray-700`}>
-            Showing {offset + 1}-{offset + rows.length} of {recordCount}
+        <div css={tw`lg:flex lg:justify-between mb-2 text-sm`}>
+          <span>
+            Showing <span css={tw`font-bold`}>{recordCount} facilities</span> in{' '}
+            {location.address}
```
Jkrzy: We should qualify this statement by including the distance of the search from the.

```Showing {recordCount} facilities within {distance} miles of {location.address}```

File: src/components/Form/FormFilters.js [hunk 4]
```diff
@@ -55,248 +38,114 @@ export class FormFilters extends Component {
...
+              legend="Medication-assisted treatment (MAT)"
+              name="mat"
+              options={filterOptions.mat}
+            />
+          </Row>
+        </RowWrapper>
+        {!isDesktop && (
+          <RowWrapper>
+            <Row>
+              <Button primary css={tw`w-full`} type="submit">
```
Jkrzy: Need to disable this when there is no `location` available

src/components/ResultsList.js
Jkrzy: Using `location.address` here also makes the rendered string modifiable as a user changes the value of the `location` input. The value here should be static and reflect the location from which the results displayed were generated.

src/components/Accordion.js
```diff
@@ -0,0 +1,44 @@
...
+      </>
+    );
+  }
+}
+
+Accordion.propTypes = {
+  children: PropTypes.node.isRequired
+};
+
+export default Accordion;
```
Jkrzy: Not seeing this component used anywhere at the moment.

src/components/Alert.js
```diff
@@ -0,0 +1,35 @@
...
+          >
+            Learn more about different types of treatment
+          </Link>
+        </p>
+      </div>
+    </div>
+  </div>
+);
+
+export default Alert;
```
Jkrzy: This component name feels misleading, what about something like `ResultsInfo`?

src/components/Card.js
```diff
@@ -1,131 +1,212 @@
...
+const StyledHeading = styled.div`
+  ${tw`px-6 py-4 bg-blue-light text-white rounded-t`}
+`;
+
+const StyledCard = tw.li`bg-white shadow border-t border-gray-lighter rounded mb-10`;
+
+const serviceIcons = {
+  TC: faPlusSquare,
+  PAY: faCreditCard
+};
```
Jkrzy: Let's include a comment here detailing the origin of the key values.

src/components/Card.js
```diff
@@ -1,131 +1,212 @@
...
+            eventLabel="Driving directions link from card thumbnail image"
+            to={`https://www.google.com/maps/dir/?api=1&destination=${encodeURI(
+              `${street1}, ${street2 && street2 + ','} ${city}, ${state} ${zip}`
+            )}`}
+            target="_blank"
+            rel="noopener noreferrer"
+          >
+            <img
+              src={`${GOOGLE_MAP_STATIC_URL}?zoom=15&size=140x113&markers=size:small%7C${latitude},${longitude}&key=${process.env.REACT_APP_GOOGLE_API_KEY}`}
+              alt="Google map"
```
Jkrzy: Let's expand this to include name of facility and differentiate from the other images which could be rendered alongside.

src/components/Form/FormFilters.js
```diff
@@ -55,248 +38,114 @@ export class FormFilters extends Component {
...
-  isDesktop: PropTypes.bool.isRequired,
-  toggleAdvancedFilters: PropTypes.func.isRequired,
-  resetAllFilters: PropTypes.func.isRequired,
-  resultsHidden: PropTypes.bool.isRequired,
-  toggleFilters: PropTypes.func.isRequired,
-  toggleResults: PropTypes.func.isRequired
+  isDesktop: PropTypes.bool.isRequired
 };
 
 const mapStateToProps = state => {
```
Jkrzy: The value of the `location` input, even when modified is being added to the store and then used as the initial value when rendered later. Even when the searched location has changed. 

To replicate: Perform a search, modify the value of `location` but don't select an autocomplete value, return to the homepage and search again for a different zip code.

src/components/Card.js
```diff
@@ -1,131 +1,212 @@
...
-  <>
-    <ul>
-      <li>
-        <span css={tw`font-semibold`}>Phone:</span>{' '}
-        <OutboundLink
+  return (
+    <div css={tw`flex flex-wrap -mb-4 -mx-6`}>
+      <div css={tw`w-full md:w-3/5 mb-4 px-6`}>
+        {renderService(services, 'TC')}
+        {renderService(services, 'PAY')}
```
Jkrzy: We can't guarantee that all returned service data will include these keys. Need to handle scenarios where these are not renderable. 

To Replicate: Search `13114`

Jkrzy:
Completed an initial pass through these changes, looking great! I made a few comments throughout, I'll help tackle/resolve each next week.

public/index.html
```diff
@@ -2,13 +2,33 @@
...
     <script src="https://maps.googleapis.com/maps/api/js?key=%REACT_APP_GOOGLE_API_KEY%&libraries=places"></script>
-    <script type="text/javascript" src="//script.crazyegg.com/pages/scripts/0083/6179.js" async="async"></script>
+    <script
+      type="text/javascript"
+      src="//script.crazyegg.com/pages/scripts/0083/6179.js"
+      async="async"
+    ></script>
   </head>
   <body>
     <noscript>You need to enable JavaScript to run this app.</noscript>
```
Jkrzy: Let's expand this to include more details about why Javascript needs to be enabled and encourage visitors to call the helpline if they're unable to access the site.

src/components/Card.js
```diff
@@ -1,140 +1,159 @@
...
           >
-            {website}
-          </OutboundLink>
-        </li>
-      )}
-    </ul>
-    <p>{renderService(services.TC)}</p>
-  </>
-);
+            plus more
```
Jkrzy: The accessibility of this link would benefit from a descriptive `aria-label` as without this link appears to be repeated throughout the results page. 

Maybe something like "Link to full service listing for {name1}", any thoughts on content @iamjolly?

src/components/Card.js
```diff
@@ -1,140 +1,159 @@
...
+            name1={name1}
+            street1={street1}
+            street2={street2}
+            city={city}
+            state={state}
+            zip={zip}
+            latitude={latitude}
+            longitude={longitude}
+            miles={miles}
+          />
```
Jkrzy: MapStatic fails pretty gracefully, not breaking layout, alt text is descriptive, and link to directions still functions. I don't think we need any additional error handling there at the moment. 

Screen Shot 2019-09-17 at 9 36 46 PM

src/components/Results/NoResults.js
```diff
@@ -0,0 +1,31 @@
...
+import tw from 'tailwind.macro';
+import 'styled-components/macro';
+import { OutboundLink } from 'react-ga';
+
+import { HELPLINE_LINK, HELPLINE_TEXT } from '../../utils/constants';
+
+const NoResults = () => (
+  <div css={tw`shadow bg-white border border-gray-lighter rounded p-6 mb-6`}>
+    <h1 css={tw`block font-bold font-heading text-2xl mb-6`}>
+      We couldn't find any providers near you.
```
Jkrzy: Content check: should this be something like "We couldn't find any providers that met your search criteria"? We know that folks don't always search near where they are located.

cc: @amandacostello

src/components/Results/NoLocation.js
```diff
@@ -0,0 +1,35 @@
...
+
+import { HELPLINE_LINK, HELPLINE_TEXT } from '../../utils/constants';
+
+const NoLocation = () => (
+  <div css={tw`shadow bg-white border border-gray-lighter rounded p-6 mb-6`}>
+    <h1 css={tw`block font-bold font-heading text-2xl mb-6`}>
+      Searching for treatment options
+    </h1>
+    <p css={tw`mb-4`}>
+      Start by entering your ZIP code. Locations within 25 miles will be shown
```
Jkrzy: Is it overly specific to include the details of how the search functionality works? Feels like yes and "geographic center" doesn't strike me as plain language.

 "Locations within 25 miles of the geographic center of your search location will be shown"

cc: @amandacostello

src/components/Map/MapStatic.js
```diff
@@ -0,0 +1,64 @@
...
+    return (
+      <>
+        <OutboundLink
+          eventLabel="Driving directions link from card thumbnail image"
+          to={`https://www.google.com/maps/dir/?api=1&destination=${encodeURI(
+            `${street1}, ${street2 && street2 + ','} ${city}, ${state} ${zip}`
+          )}`}
+          target="_blank"
+          rel="noopener noreferrer"
+        >
```
Jkrzy: Let's include a descriptive label for this link reflecting the target of directions to the facility.

public/index.html
hursey013: @amandacostello would you be able to provide some content for this scenario - I'm guessing it could be similar to our error page. Basically this message will appear if someone does not have JavaScript enabled in their browser, in which case React will not function and the application will not be loaded. In lieu of the website we can provide some simple text/links to help folks to know what to do next.

src/components/Card.js
iamjolly: @Jkrzy:

If there is a desire to keep the "plus more" as the visible link text (which is fine in context), then I would recommend appending a visually-hidden span within the link with content something like:

`plus more <span class="vh">services for {name1}</span>`

Alternately, an `aria-label` for the link would override the "plus more" with whatever you apply to it, so in that case, "All services for {name1}" or something to that effect would be good.

hursey013: Great, just pushed a commit with an `aria-label`, thanks!

public/index.html
amandacostello: Text for folks who don't have JavaScript turned on

*Please turn on JavaScript for your browser*
This site uses JavaScript, a programming language that runs on your web browser. 
You're seeing this message because JavaScript is currently *turned off* or disabled for your browser. This site uses JavaScript to search for and show treatment facilities. 

To use this site, please turn on or enable JavaScript in your browser's settings. This is often found in the _Security_ or _Privacy_ settings. 

You can always call the HelpLine to talk to a person who can help you find treatment options. They're available at 1-800-662-HELP (4357).

PR #326, opened 2019-09-17
Made content changes and added filter

allisonnorman:
- Made many minor content changes (some typos, a couple word/label changes) based on the feedback from TTS folks today
- Added the AH filter to special groups and services for deaf and hard of hearing


src/components/Results/index.js
```diff
@@ -110,7 +110,7 @@ export class Results extends Component {
           to="/content/treatment-options/types-of-treatment"
           css={tw`underline text-gray-darkest`}
         >
-          Learn more about different types of treatment
+          Learn more about different types of treatment.
```
hursey013: I made this one already, there is a period just outside of the `</Link>` below

allisonnorman: Good catch. That's the one thing I couldn't see locally. I'll push a change removing this.

allisonnorman:
@hursey013 - Pushed the latest change which includes the update to the mobile H1 text which you mentioned you would help make more mobile-friendly line break wise. I'm seeing a conflicting file message -- let me know if I can help with that.

PR #333, opened 2019-09-19
[WIP] Implement DAP integration

Jkrzy:
Let's participate in https://digital.gov/services/dap/.

This PR is an initial implementation of pulling in the DAP code alongside our existing GA setup and generating virtual page views for DAP. 

For now, we'll leave our custom events in the SAMSHA prototyping GA space.


Jkrzy:
Herein we send virtual pageviews to DAP on arrival to the site and for each navigation action a user takes.

We're preserving the existing functionality of sending pageviews as well as custom events to the SAMHSA/18F property created specifically for this effort

src/store.js
```diff
@@ -17,6 +17,10 @@ if (process.env.NODE_ENV !== 'production') {
     collapsed: true
   });
   middleware.push(logger);
+
+  if (process.env.REACT_APP_LOCAL_ANALYTICS_TESTING) {
+    middleware.push(analytics);
+  }
```
hursey013: Looks like we can drop this

src/index.js
```diff
@@ -17,11 +17,14 @@ if (process.env.NODE_ENV !== 'production') {
...
-    : process.env.REACT_APP_DEV_GA_TRACKING_ID
-);
+ReactGA.initialize([
+  {
+    trackingId:
+      process.env.REACT_APP_BRANCH === process.env.REACT_APP_PROD_BRANCH
+        ? process.env.REACT_APP_PROD_GA_TRACKING_ID
+        : process.env.REACT_APP_DEV_GA_TRACKING_ID
+  }
+]);
```
hursey013: Is passing an array necessary here since we will always only have one tracking ID?

src/middleware/analytics.js
```diff
@@ -10,22 +10,33 @@ const trackEvent = action => {
...
+  }, 0);
 };
 
 let currentPage = '';
 
 export const analytics = store => next => action => {
-  if (action.type === '@@router/LOCATION_CHANGE') {
+  if (
+    action.type === '@@router/LOCATION_CHANGE' &&
+    ['PUSH', 'POP'].includes(action.payload.action)
```
hursey013: It looks like we are only `PUSH` actions in `trackPage()` so do we need to include `POP` here?

Jkrzy: Implemented this way so that ReactGA sends pageviews on both PUSH & POP (first arrival).

The DAP script sends an initial arrival pageview automatically when the file is retrieved and initialized

Jkrzy: Excluding "REPLACE" location changes to accommodate the current functionality on the content pages.

src/index.js
Jkrzy: Ah good catch, leftover from dev work

src/middleware/analytics.js
davemcorwin: as part of #411 I am doing a a little refactoring that will remove the Redirect, so should I/we remove the extra condition when that happens?

src/store.js
Jkrzy: Would like to leave to facilitate future local dev work

Jkrzy:
@davemcorwin, ready for your 👀. Made a few tweaks stemming from @hursey013's comments.

src/middleware/analytics.js
```diff
@@ -10,22 +10,33 @@ const trackEvent = action => {
...
-const trackPage = page => {
+const trackPage = (action, page) => {
   ReactGA.set({
     page,
     ...options
   });
-  ReactGA.pageview(page);
+
+  // Allow page title to update before sending
+  setTimeout(() => {
```
davemcorwin: Do you have a sense of how reliable this will be and/or how much we care?

I think it's possible that depending on the changes and the device this could run before the title is updated. I think Brian mentioned that you could use the `onChangeClientState` callback provided by [`react-helmet`](https://github.com/nfl/react-helmet#reference-guide) (which I believe runs during the `componentDidMount` lifecycle event) to guarantee that the updated document title is available when this fires.

Jkrzy:
Thanks @hursey013 and @davemcorwin, this was much easier w/ `react-helmet`.

Remaining issue with this approach, we're double counting and generating 2 DAP pageviews on arrival. One when the DAP script is loaded and one when `react-helmet` fires up.

PR #337, opened 2019-09-19
Couple more content tweaks

allisonnorman:


allisonnorman:
This is pulling in other changes (probably because I forgot to fetch). Going to close and recreate.

PR #338, opened 2019-09-19
Couple more content tweaks

allisonnorman:


allisonnorman:
@hursey013 - can you tell what's happening? I made two changes but it looks like a lot more came through. I'm assuming it's a formatting thing that's happening? If it's on my end, no worries and I can just make the changes in github tomorrow.

hursey013:
Did you make the change locally? If so, it could be `prettier` doing some autoformatting?

PR #342, opened 2019-09-20
Update footer

allisonnorman:
- Revised wording to include "emotional distress" in Lifeline callout
- Telephone linked Lifeline and SAMHSA #s in footer


hursey013:
Used `OutboundLink` component so clicks are registered in GA as events.

PR #348, opened 2019-09-23
Remove 'all types of treatment' option from MAT section

### LINKED ISSUE #293: Remove "all types of treatment" from MAT dropdown
## Issue: 
Feedback on the beta site shows possible confusion with the MAT dropdown list. I think this might be a search blocker.

> if you expand the filters you have to choose a MAT. This may not be appropriate for all persons seeking treatment. I think the drop down should also have an N/A option.

Screen Shot 2019-09-06 at 8 25 04 AM
**UPDATE FOR MVP**: Remove "all types of treatment" option from MAT section

## Suggestion:
Add an N/A button
- this gives the visitor the option to do an action that gives them more confidence to complete their search
- Because of the limitations of the OR logic with the filters I would have the filtering mechanism show the same results for All Treatments

allisonnorman:
We decided to remove the "all types of treatment" default option from the MAT filter section. It was confusing to people in usability tests. 

Resolves #293

### DIFF
--- a/src/utils/filters.js
+++ b/src/utils/filters.js
@@ -17,6 +17,12 @@ export const distance = [
 
 export const mat = [
   {
+    value: '',
+    label: 'Show all facilities',
+    description:
+      'Includes facilities that do not offer medication-assisted treatment'
+  },
+  {
     value: 'METH',
     label: 'Methadone',
     description:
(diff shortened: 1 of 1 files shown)


allisonnorman:
Oh yeah, that's tricky. I looked at it on the preview branch but didn't click around enough to see this happens. We have an alternate option that the team discussed that I could put in place, which is to add back in the "all" but use explainer text. I'll do that and push a change today. Thanks for catching this!

allisonnorman:
@hursey013 - Updated this to add back in the default option.

PR #353, opened 2019-09-24
Adds button to reset filters to desktop search

### LINKED ISSUE #349: Add "reset filters" back in
I think we list this functionality as the design has been implemented. 

Here was the original PR (I believe) implementing this originally: #27

davemcorwin:
Resolves #349 

Adds a button to reset the form to default values.

This is pending OK from design @austinhernandez @bpdesigns , but I wanted to create the PR for visibility and make sure the preview deploy is available.


src/components/Form/FormFilters.js
```diff
@@ -58,14 +58,19 @@ export class FormFilters extends Component {
...
       <div css={tw`bg-teal-lighter rounded shadow border border-gray-light`}>
         {isDesktop && (
-          <h2 css={tw`text-2xl font-heading font-bold p-4 shadow`}>
-            Refine search results
-          </h2>
+          <div css={tw`p-4 shadow`}>
+            <h2 css={tw`text-2xl font-heading font-bold`}>
+              Refine search results
+            </h2>
+            <Button link onClick={resetForm} id="reset-form">
```
davemcorwin: Added the `id` attribute here solely for the purposes of identifying the element during testing. I don't know if there is a preferred way of doing this for this project or not. I have also used `data` attributes for this purpose in the past as they can be more clear in identifying WHY the attribute is present. I'm open to alternative solutions.

src/components/Form/FormFilters.js
```diff
@@ -183,7 +189,14 @@ const mapStateToProps = state => {
   };
 };
 
-export default connect(mapStateToProps)(
+const mapDispatchToProps = dispatch => ({
+  resetForm: dispatch(reset('filters'))
```
hursey013: I'm having trouble getting this dispatched as is when the button is clicked. I've set this up as a function in other places which seems to do the trick:
```
  resetForm() {
    dispatch(reset('filters'));
  }
```

src/components/Form/FormFilters.js
hursey013: I do not have a strong opinion - I've been using class names, or because we are using styled-components that will often generate a unique name to target, but `id` works as well.

davemcorwin: Replaced with a class for consistency.

src/components/Form/FormFilters.js
davemcorwin: This was my bad, forgot to make it a new function...

austinhernandez:
This is good for MVP

Issue #344, opened 2019-09-23
Before FindTreatment.gov launch checklist


allisonnorman:
**Set up findtreatment.gov**
- [x] Set up Crazy Egg snapshot #472
- [x] Disable or add warning banner to findtreatmentbeta.18f.gov #476 
- [x] Preview google sitelinks in Google console (https://support.google.com/webmasters/answer/47334?hl=en)
- [x] Update `REACT_APP_SITE_DOMAIN` env var to be `https://findtreatment.gov` #474

**Make changes to prototype** 
- [x] Update the url referenced in Readme.md (line 7) to `https://findtreatment.gov`
- [x] Update header and footer with new feedback form links (use: https://forms.gle/ceUB7jd2GULXdQj96) #479
- [x] Change name of website in logo and elsewhere (browser tab name, etc) to FindTreatment.gov (instead of TreatmentFinder)  #478
- [x] Remove robots.tx (reverse #354)
- [x] Remove link to findtreatment.samhsa.gov in beta banner (it's now included in footer) #473 
- [x] Point site at new domain (per WH approval re: timing). Domain has been fully registered (#272). Double check that routing is working properly (no issues going directly to links)

**Verify**
- [x] Check social sharing works (Twitter, FB, etc) #477 

**Communicate**
- [x] Give heads up to Federalist team about launch date. Note: Britta said this is important as Federalist is classified as low impact and they should know when we expect a higher level of traffic
- [x] Notify DAP of new domain participation

hursey013:
Added a bullet for updating the `REACT_APP_SITE_DOMAIN` variable - this is used in the meta tags in order to provide a full absolute URL (https://github.com/18F/samhsa-prototype/pull/437).

hursey013:
Once findtreatment.gov is launched we should either disable findtreatmentbeta.18f.gov completely, or add a banner to it making it clear that it is a testing site and folks should be using the live site instead.

PR #354, opened 2019-09-24
Prevent search engine indexing

hursey013:
Add robots.txt to prevent search engine indexing of beta site.


allisonnorman:
@hursey013 - Just to confirm, there's no harm in pushing this now (as we don't mind if it's not indexed today), and it's easily reversible before launch? If so, we're good to merge! Thank you!

hursey013:
@allisonnorman yep, using a robots.txt is a pretty common approach, so no harm in having it. Once we remove it (or adjust it to allow indexing) it will be scanned as per usual by search engines.

hursey013:
@allisonnorman I'm not sure where we are tracking it, but removing the `robots.txt` should be added to our launch checklist.

allisonnorman:
Tracking here: #344

PR #357, opened 2019-09-24
Filter enhancements

### LINKED ISSUE #330: Filter enhancements
Screen Shot 2019-09-18 at 12 43 52 PM

- [ ] Color for helper text #2C3747

- [ ] Align radio button middle

- [ ] Increase the size of the radio button

- [ ] Active radio button color link blue

- [ ] A little more space (margin-bottom: 1rem;) between `<label>` elements

davemcorwin:
Resolves #330 

I believe this addresses everything but changing the color of the radio inputs. Per my comment on the issue, I am guessing that implementing custom inputs JUST to change the color is probably out of scope for the MVP...

Changing the size of the radio inputs on mobile is done with `transform: scale(1.25)`. This appears OK when viewing the "mobile" site in Mac Chrome, but needs some browser testing, especially on mobile, to make sure this is working as intended.

### DIFF
--- a/src/components/Input/Label.js
+++ b/src/components/Input/Label.js
@@ -5,14 +5,16 @@ import { PropTypes } from 'prop-types';
 import { Link } from 'react-router-dom';
 
 const StyledLabel = styled.label`
-  ${tw`block mb-2`}
+  ${tw`block mb-4`}
 
   ${props => props.inline && tw`flex`}
 
   span {
     ${tw`block font-bold font-heading uppercase text-sm mb-2`}
 
-    ${props => props.inline && tw`font-sans normal-case mb-0`}
+    ${props =>
+      props.inline &&
+      tw`font-sans normal-case leading-none mb-1 sm:leading-normal sm:-mt-px`}
 
     ${props => props.large && tw`normal-case text-2xl md:text-3xl`}
   }
@@ -28,14 +30,10 @@ const Label = props => {
       {props.inline ? (
         <StyledLabel {...props}>
           {props.children}
-          {props.description ? (
-            <div css={tw`-mt-px`}>
-              <span>{props.labelText}</span>
-              <p>{props.description}</p>
-            </div>
-          ) : (
+          <div css={tw``}>
             <span>{props.labelText}</span>
-          )}
+            {props.description && <p>{props.description}</p>}
+          </div>
         </StyledLabel>
       ) : (
         <StyledLabel {...props}>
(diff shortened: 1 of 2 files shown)


src/components/Input/Label.js
```diff
@@ -28,14 +30,10 @@ const Label = props => {
...
-          {props.description ? (
-            <div css={tw`-mt-px`}>
-              <span>{props.labelText}</span>
-              <p>{props.description}</p>
-            </div>
-          ) : (
+          <div>
             <span>{props.labelText}</span>
-          )}
+            {props.description && <p>{props.description}</p>}
```
hursey013: Much cleaner ✨

src/components/Input/Label.js
```diff
@@ -28,14 +30,10 @@ const Label = props => {
       {props.inline ? (
         <StyledLabel {...props}>
           {props.children}
-          {props.description ? (
-            <div css={tw`-mt-px`}>
-              <span>{props.labelText}</span>
-              <p>{props.description}</p>
-            </div>
-          ) : (
+          <div>
```
hursey013: This could be a React Fragment `<>`, since it's not necessary style wise.

src/components/Input/Label.js
```diff
@@ -5,20 +5,22 @@ import { PropTypes } from 'prop-types';
...
 
   ${props => props.inline && tw`flex`}
 
   span {
     ${tw`block font-bold font-heading uppercase text-sm mb-2`}
 
-    ${props => props.inline && tw`font-sans normal-case mb-0`}
+    ${props =>
+      props.inline &&
+      tw`font-sans normal-case leading-none mb-1 sm:leading-normal sm:-mt-px`}
```
hursey013: Just want to double check the intent here, is it to have `leading-none` on mobile and `leading-normal` on screens > than mobile?

davemcorwin: yes, bc otherwise it was getting wonky on tablet sizes. I was attempting to align the radio button with the text by matching the line heights but I wasn’t entirely successful.

src/components/Input/Label.js
davemcorwin: I think in this case it IS necessary. When `props.inline == true` the parent `label` element is `display: flex` and we want the `span` and the `p`, if present, to be considered a single flex item.

hursey013: Yep, my mistake, carry on!

PR #361, opened 2019-09-25
Add disabled attr to button when disabled

### LINKED ISSUE #307: Homepage - find treatment validation missing from empty submission
### Problem Description
The find treatment form does nothing when a user tries to submit the form with a blank/empty search field. No validations seem to happen and no error message is displayed. This can be a confusing experience, especially for those with cognitive and sight-related disabilities. 

samhsa-locate-submit-press-without-feedback

### Recommended Solution(s)
Provide the same/similar error validation and notification experience as for those who enter invalid data in the field. See screenshot below of how this notification should also be altered for a better user experience, particularly for those with disabilities (will file related, separate issue, too).

samhsa-locate-validation-msg

Alternately, the submit button could be `disabled` via HTML attribute until valid (or any) data is in the search `input` field.

davemcorwin:
Resolves #307 

Adds the `disabled` attribute to `Button`s when they are disabled.

@iamjolly I don't think this is your preferred approach, but this is MUCH quicker right now than showing the error message on submit of no data. If we still want to enhance this to use other behavior, perhaps we can roll it into #308 and reprioritize.

### DIFF
--- a/src/components/Input/Button.js
+++ b/src/components/Input/Button.js
@@ -6,25 +6,13 @@ import { PropTypes } from 'prop-types';
 import { theme } from '../../tailwind.js';
 
 const Button = styled(
-  ({
-    as,
-    children,
-    primary,
-    secondary,
-    outline,
-    gradient,
-    disabled,
-    link,
-    ...rest
-  }) =>
+  ({ as, children, primary, secondary, outline, gradient, link, ...rest }) =>
     as ? (
-      <Button as={as} disabled={disabled} {...rest}>
+      <Button as={as} {...rest}>
         {children}
       </Button>
     ) : (
-      <button disabled={disabled} {...rest}>
-        {children}
-      </button>
+      <button {...rest}>{children}</button>
     )
 )`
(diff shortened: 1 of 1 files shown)


hursey013:
This absolutely works as is, and I'll defer to you on your preference, but I realized this was actually my mistake in pulling out the `disabled` prop and not passing it through to the component. I was having issues with some of these different style props getting passed through to the DOM so I removing them and passing on the rest, but in this case I didn't need to pull disabled out since that is a valid DOM attribute. So basically, the alternative approach would just be to remove `disabled` here https://github.com/18F/samhsa-prototype/blob/1efc44a2b0bdfbd75f9e0d062feb18c9add2e15e/src/components/Input/Button.js#L16 and then it wouldnt need to be added to each render since it'll get passed through with `...rest`.

iamjolly:
this all seems totally fine with setting and removing `disabled` on the `button` based on empty vs bad vs valid data in the `input`. Thanks for tackling this one!

Issue #218, opened 2019-08-09
Improve integrated content UX


allisonnorman:
**UPDATE FOR MVP**: 

- [x] Tooltips at filter category level on results page (payment options, as opposed to each option)) - @amandacostello to share tooltip content for each section
- [x] Payment tooltip on details page (not on design, but we can reuse same wording for results page tooltip if @amandacostello confirms)
- [x] Add an "i" icon to "before you call" banner"
- [ ] Smaller cards on details page (transportation is or is not available, and insurance coverage  box) - See design for details page
- [x] @amandacostello - You said you wanted to make changes to the "before you visit" section on the results page. We're not sure if this should be collapsed by default or not. Once you determine the changes that need to be made, can we align on whether this should be collapsed by default?
- [x] (Maybe, depending on above) Collapse "before you go" section on results page by default
______________

Currently, we have links throughout the site that direct people to info content to find out more if they're unfamiliar with the task at hand. 

For example: linking to treatment options near the drop down for type of care ('not sure?'), "before you call" link on the facility cards and details pages, etc. 

Right now, we direct people off of the search page which isn't ideal. We think this could be improved in a way that doesn't disrupt the person's search.

Some ideas: 
- Tooltip
- Modal
- What else?

austinhernandez:
## Tooltips

### Mobile tooltip:

![Tooltip - mobile](https://user-images.githubusercontent.com/23194720/64040387-e73b5100-cb2a-11e9-8c90-ab97004ad773.png)

Interaction pattern:
[Similar to the bottom pop up in mobile view on this page.](https://www.zocdoc.com/doctor/hari-polavarapu-md-7897?LocIdent=97738&dr_specialty=153&insuranceCarrier=324&insurancePlan=2778&isNewPatient=true&reason_visit=75
)


### Desktop tooltip:

![Results page tooltip](https://user-images.githubusercontent.com/23194720/64043068-63d12e00-cb31-11e9-9dba-03547a6154ec.png)

Interaction pattern:
I'm thinking this is more of a click than hover situation because we should link to the "What to Expect page for the lonnnngggggg explanation. Open to suggestions on this, though.

bpdesigns:
@allisonnorman @austinhernandez was this determined to be out of scope? I think the tooltips that @austinhernandez designed look great.

allisonnorman:
[...]

PR #362, opened 2019-09-25
Add transportation and payment callouts to detail page

hursey013:
Fixes https://github.com/18F/samhsa-prototype/issues/218

This is the last item on the checklist for the issue above, however I believe there might be some additional tasks that should be pull out of the comments of that issue and made into new issues prior to closing.

This adds a conditional message to the Location section of the details page depending on whether the facility offers transportation or not. It also adds a static warning to the payment section to double check with the provider.

Preview:
Screen Shot 2019-09-25 at 2 52 22 PM

@davemcorwin this also includes some initial steps for breaking the `Details` components into smaller chunks.


src/components/Details/DetailsLocation.js
```diff
@@ -0,0 +1,103 @@
...
+    street2,
+    city,
+    state,
+    zip,
+    services,
+    phone
+  } = data;
+
+  const offersTransportation = (
+    ((services || {}).AS || {}).values || []
```
davemcorwin: ugh, but the future is bright!
https://github.com/tc39/proposal-optional-chaining

hursey013: Yeah, I'm not sure what the best patten is here, this method feels the most reliable, but yeah, not pretty 😦

PR #366, opened 2019-09-25
Update address styles

### LINKED ISSUE #347: Make the address on results page a link
- [ ] Change the color of the address to a `gray dark` 
- [ ] Make the address a clickable link
- [ ] Include the container and miles away also clickable
- [ ] Add a tiny bit (.1rem) of spacing between the phone number, website, and address

Screen Shot 2019-09-23 at 3 29 04 PM

davemcorwin:
Resolves #347 

Updates the address styles and links per the issue. Some refactoring to consolidate the address formatting and generation of external google maps link.

Screen Shot 2019-09-25 at 4 41 54 PM
Screen Shot 2019-09-25 at 4 42 05 PM

### DIFF
--- a/src/components/Card.js
+++ b/src/components/Card.js
@@ -7,7 +7,12 @@ import { OutboundLink } from 'react-ga';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { faCheck, faPhone } from '@fortawesome/free-solid-svg-icons';
 
-import { linkToFacility, removeHttp } from '../utils/misc';
+import {
+  googleMapUrl,
+  linkToFacility,
+  removeHttp,
+  formatAddress
+} from '../utils/misc';
 
 import MapStatic from './Map/MapStatic';
 
@@ -71,6 +76,14 @@ const Card = props => {
     zip
   } = props;
 
+  const address = {
+    street1,
+    street2,
+    city,
+    state,
+    zip
+  };
+
   return (
     <li css={tw`bg-white shadow border-t border-gray-lighter rounded mb-10`}>
       <div css={tw`px-6 py-4 bg-blue-light text-white rounded-t`}>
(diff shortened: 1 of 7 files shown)


src/utils/misc.test.js
```diff
@@ -33,3 +34,56 @@ describe('linkToFacility()', () => {
...
+  });
+
+  it('adds a line break when `lineBreak` is true and street2', () => {
+    expect(misc.formatAddress(addressWithStreet2, true)).toEqual(
+      '19201 120th Avenue NE, Suite 108,\nBothell, WA 98011'
+    );
+  });
+});
+
+describe('googleMapUrl()', () => {
```
davemcorwin: this test is fairly pointless...

src/components/Card.js
```diff
@@ -71,6 +76,14 @@ const Card = props => {
     zip
   } = props;
 
+  const address = {
```
hursey013: This gave me an idea - what do you think about creating this address object in the facilities/facility reducer so that it comes to the component already grouped together?

src/utils/misc.js
```diff
@@ -36,3 +37,17 @@ export const linkToFacility = ({ frid, latitude, longitude }) => {
     search: qs.stringify({ sAddr: `${latitude}, ${longitude}` })
   };
 };
+
+export const formatAddress = (
+  { street1, street2, city, state, zip },
+  lineBreak = false
+) => {
+  return `${street1},${street2 ? ` ${street2},` : ''}${
+    lineBreak ? '\n' : ' '
```
hursey013: I didn't realize you could use `\n` link breaks like this. Is there any advantage/disadvantage over using an html `<br />` like:
```
<>
{street1}, {street2 && street2 + ','}<br />
{city}, {state} {zip}
</>
```

davemcorwin: We can do it either way, I don't have a strong preference. I used `\n` so the method returns a plain string instead of an html string, but this necessitated the use of css `whitespace: pre-line`.

src/components/Card.js
davemcorwin: This makes a lot of sense to do, but I'm going to defer it for now. Created #374 to track it for future work.

src/utils/misc.js
hursey013: Good to know!

PR #367, opened 2019-09-26
Ådd GitHub link to footer

### LINKED ISSUE #254: Add GitHub link to prototype
To support developing in the open, and to open up technical feedback/contributions.

davemcorwin:
Resolves #254 

Here is an initial take of adding the link. Added @amandacostello as a review for feedback on the actual content.

Screen Shot 2019-09-25 at 5 12 57 PM

### DIFF
--- a/src/components/Footer.js
+++ b/src/components/Footer.js
@@ -123,7 +123,7 @@ const Footer = () => {
                   <OutboundLink
                     to="https://github.com/18F/samhsa-prototype"
                     eventLabel="Browse our source code"
-                    css={tw`flex -mt-px`}
+                    css={tw`flex`}
                   >
                     <FontAwesomeIcon
                       icon={faGithub}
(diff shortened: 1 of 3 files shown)


src/components/Footer.js
```diff
@@ -117,6 +119,19 @@ const Footer = () => {
                     Submit feedback
                   </OutboundLink>
                 </li>
+                <li>
+                  <OutboundLink
+                    to="https://github.com/18F/samhsa-prototype"
+                    eventLabel="Browse our source code"
+                    css={tw`flex`}
```
hursey013: Since you are using `flex`, `items-center` can handle horizontal alignment instead of needing to adjust the top margin of the icon, but the difference is likely not discernible in this case.

hursey013:
I think for the people that look for/understand open source, having a `View code on GitHub` is a nice touch, and since it's tucked away in the footer I wouldn't expect it to distract or confuse folks that aren't familiar with it.

davemcorwin:
Updated content to `View code on Github`

PR #370, opened 2019-09-26
Add horizontal rules and location distance to details page

hursey013:
**Before:**
Screen Shot 2019-09-26 at 10 59 06 AM

**After:**
Screen Shot 2019-09-26 at 10 59 34 AM

Also moved the formatting of the distance to a helper function.


hursey013:
@davemcorwin added an additional test for a more typical mile value - any others you had in mind?

davemcorwin:
Thanks for adding tests! Maybe where `miles` is `null` or `undefined` since that case is handled in the method, but the PR gtg whenever.

PR #372, opened 2019-09-26
Add small animation to mobile button when results are being retrieved

hursey013:
Smooths the transition when populating the button with the number of results on mobile.

**Preview:**
Screen Shot 2019-09-26 at 11 33 38 AM


davemcorwin:
Implementation looks 💯 . I defer to @austinhernandez but maybe consider replacing all of the text with the spinner instead of the just the number of results?

PR #375, opened 2019-09-26
Focus on the `h1` when an error occurs

### LINKED ISSUE #316: Error page - focus management
Related to #314.

### Problem Description
When the page/screen loads the error message (happens when treatment facility results page experiences an issue), no indication is provided to blind or very low vision users about the error or new content on the screen.

samhsa-error-h2-focus

### Recommended Solution
For this error route, consider setting focus on the `h2` as this will announce the error for screen reader users, and will likely focus the screen on that content for low vision users using extreme magnification. That way, the heading, then the following paragraph and CTA to call the toll-free number will be directly available for those users.

davemcorwin:
Resolves #316 

Screen Shot 2019-09-26 at 11 47 40 AM
Screen Shot 2019-09-26 at 11 47 59 AM

### DIFF
--- a/src/components/Error.js
+++ b/src/components/Error.js
@@ -23,13 +23,13 @@ class Error extends Component {
         <Helmet>
           <title>Error</title>
         </Helmet>
-        <h2
+        <h1
           css={tw`text-3xl font-bold mb-10`}
           ref={this.headerRef}
           tabIndex="-1"
         >
           Something's wrong on our side.
-        </h2>
+        </h1>
         <div css={tw`max-w-xl mx-auto`}>
           <p css={tw`text-xl mb-6`}>
             We're having technical issues right now. The problem should be fixed
--- a/src/components/Error.js
+++ b/src/components/Error.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { Component } from 'react';
 import tw from 'tailwind.macro';
 import 'styled-components/macro';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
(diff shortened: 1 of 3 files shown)


hursey013:
@davemcorwin while we're at it, this should probably be a `h1`, what do you think?

davemcorwin:
yeah @hursey013 there is no `h1` on the page

hursey013:
Not sure why but I'm not getting the header focused when I'm looking at it locally? I'm testing it by just going to `http://localhost:3000/foo`.

hursey013:
Ugh, nvm, my fault. This is the error page, and I'm looking at the not found page. I think these updates should also be applied there as well - it's the `NoMatch` component.

davemcorwin:
oohhhh, great catch! Different components. `Error` and `NoMatch`. Yes, they are pretty much the same component though, I will consolidate.

davemcorwin:
@hursey013 Thanks for the feedback, give it another looksee!

src/components/Error.js
```diff
@@ -10,29 +11,59 @@ import { HELPLINE_LINK, HELPLINE_TEXT } from '../utils/constants';
...
+          css={tw`text-3xl font-bold mb-10`}
+          ref={this.headerRef}
+          tabIndex="-1"
+        >
+          {headerText}
+        </h1>
+        <div css={tw`max-w-xl mx-auto`}>
+          <p css={tw`text-xl mb-6`}>{description}</p>
+          <Button
+            eventLabel={helpLabel}
```
hursey013: If you want to shave off one prop, we could do:
```suggestion
            eventLabel={`Helpline link from ${title} page`}
```

davemcorwin: This would change `Helpline link from 404` to `Helpline link from Not found` for the `NoMatch` component, is that ok? I can also fix the capitalization...

hursey013: Yeah I think that's ok - these are just the labels that are associated with custom events we've defined in google analytics. Since we haven't launched yet I think it's ok to change them. Post launch we may want to be more systematic about updating them.

PR #377, opened 2019-09-26
Add MH link to top nav on desktop

hursey013:
We removed the MH link in the top header to save some space, but it feels odd from an information architecture perspective that it's not there, but is in the mobile menu, on the homepage, and in the sidebar on the content pages.

This adds the link back in, and attempts to rename the link to avoid having two `Understanding...` links in a row.


hursey013:
Would love your thoughts on this @amandacostello

allisonnorman:
I really like this. We would need to double check any links pointing to the /understanding-mental-health. @amandacostello - once you take a look, we should mention it to Brooklyn as well to make sure she's okay with this change

hursey013:
@allisonnorman I did a search for `/understanding-mental-health` links and didn't find any so I think we are ok with regards to that.

allisonnorman:
Updated this PR to bring back in 'understanding mental health' per Brooklyn. @hursey013 - Ready to merge if it looks good to you.

hursey013:
@allisonnorman Since the links on mobile and desktop are now in sync, I consolidated the links into a single array vs having a separate array for each one.

PR #381, opened 2019-09-26
Remove Google Maps IFrame

### LINKED ISSUE #186: Google maps iframe accessibility lacking name attr and hidden focus
The `iframe` in which google maps is rendered on the /results page does not have a title or name attribute. Can this be resolved on our end? https://accessibility.18f.gov/iframes/

When navigating via keyboard, the iframe is also selectable although focus is hidden. 

_Originally posted by @Jkrzy in https://github.com/18F/samhsa-prototype/issues/184#issuecomment-518390655_

davemcorwin:
Resolves #186 maybe?

So this is a bit weird, but it doesn't look like the IFrame is being used, so what if we just remove it???? Looks like everything still works...


hursey013:
I don't know about this one. I wonder if the iframe is used in some sort of fallback situation? I think the other issue with the map (with and without the iframe removed) is it does not have a visible focus ring when it has focus.

I'm sort of at a loss on how to best address this - haven't found a lot of guidance when I've searched for workarounds.

davemcorwin:
We could evaluate a different Google Maps Component library, that is what is adding a bunch of extra divs...

davemcorwin:
@hursey013 @iamjolly This works visually, but is a bit of a hack.

I'm using the `appearance` css property to achieve the browser default outline, I don't know the effect of this on all browsers, but it seems to work on on Chrome, Firefox, Safari.

Jkrzy:
Since the original issue was posted, we've moved to displaying the dynamic map only on individual facility pages. Less exposure for this potential problem.

The fix proposed here is movement in the right direction but doesn't fully resolve the issue.

There's another `<div>` in there that takes focus, with no visual indication, on keyboard navigation and allows users to pan the map with keyboard controls.

I think this is a reasonable intermediate measure that improves the existing experience.

PR #384, opened 2019-09-27
Add `return to top` link

### LINKED ISSUE #319: back to top button on mobile
Add a "back to top" button at the bottom of all mobile pages.

We have some looong pages and it would aid navigation.

davemcorwin:
Resolves #319 
Screen Shot 2019-09-27 at 9 23 06 AM

### DIFF
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -7,6 +7,7 @@ import withSizes from 'react-sizes';
 import { Helmet } from 'react-helmet';
 
 import { theme } from '../tailwind.js';
+import { TOP_ID } from '../utils/constants';
 
 import ScreenContext from './ScreenContext';
 import GlobalStyle from './GlobalStyle';
@@ -27,7 +28,7 @@ class App extends Component {
           defaultTitle={process.env.REACT_APP_SITE_TITLE}
         />
         <GlobalStyle />
-        <div id="top" css={tw`overflow-hidden`}>
+        <div id={TOP_ID} css={tw`overflow-hidden`} tabIndex="-1">
           <Header />
           <main role="main">
             <Switch>
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -42,11 +42,6 @@ class App extends Component {
               <Route component={NoMatch} />
             </Switch>
           </main>
-          <div className="container" css={tw`mb-4 sm:hidden`}>
-            <a href="#top" aria-label="Return to top" css={tw`text-blue`}>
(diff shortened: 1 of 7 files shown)


src/components/App.js
```diff
@@ -42,6 +42,11 @@ class App extends Component {
               <Route component={NoMatch} />
             </Switch>
           </main>
+          <div className="container" css={tw`mb-4 sm:hidden`}>
+            <a href="#top" aria-label="Return to top" css={tw`text-blue`}>
```
davemcorwin: @hursey013 is there a better way of doing this with React Router?

davemcorwin:
@iamjolly Does this need an announcement for sr?

iamjolly:
@davemcorwin - Nope, this doesn't need any announcement for SRs since it's linking to a named anchor on the page. Just ensure the destination receives focus, and it's good to go without any announcement.

hursey013:
Couple of thoughts:
- Would there be any disadvantage to displaying this on desktop as well?
- Any reason to not include this in `<App />` right about the `<Footer />` so it's available globally?
- Last thing would be just to address @iamjolly suggestion of changing the focus as well - perhaps to the `h1` or maybe `<main>`?

src/components/App.js
hursey013: I think this is fine. We have the `HashLink` component available - https://github.com/18F/samhsa-prototype/blob/master/src/components/Home.js#L132-L140 - but I'm not really sure there is any advantage to using that since we don't need smooth scrolling and the link is targeting an anchor link on the same page.

davemcorwin:
Thanks for all the feedback! Updated to set focus on the top element after navigation AND use the `HashLink` component for consistency.

1. I am navigating to the parent `div` instead of `main` since that navigated to just below the header (the image was the topmost visible item) and that seemed awkward, but open to changing it.

2. I am including the component in each page individually because the backgrounds are different colors (Ex. white on Home page, gray on Results), so I can't match it otherwise. (I suppose I could make it explicitly dependent on the route, but that seems awkwardly coupled and error prone)

3. Adding to desktop makes sense, especially on the Results page when there are results and on some of the content pages. However, where should it go in these cases? For Example, It looks awkward all the way on the left on the Results page, maybe in the middle of the page?

Screen Shot 2019-09-27 at 2 49 45 PM

davemcorwin:
Created #394 to track possible addition to Desktop view, merging this for now.

PR #390, opened 2019-09-27
Details page visual design improvements

hursey013:
Additional design improvements with a focus on responsive design.

Before:
Screen Shot 2019-09-27 at 3 24 44 PM

After:
Screen Shot 2019-09-27 at 3 24 20 PM

### DIFF
--- a/src/components/Details/DetailsLocation.js
+++ b/src/components/Details/DetailsLocation.js
@@ -5,14 +5,15 @@ import { PropTypes } from 'prop-types';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import {
   faCar,
-  faExclamationTriangle
+  faExclamationTriangle,
+  faExternalLinkAlt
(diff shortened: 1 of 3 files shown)


hursey013:
@austinhernandez just to clarify, the second screenshot is the updated one and the button says `Get directions` in it - that is what was in the mockup, is that correct?

Also, yes the distance is are implemented. However, when you are directly linking to a facility (like someone sent you the link), since there is no context on what your location is it doesn't display a distance.

hursey013:
@austinhernandez here's the preview link if you want to check out the real thing instead of the screenshot: https://cg-efcd7c90-b365-4336-8b87-6174f1975c99.app.cloud.gov/preview/18f/samhsa-prototype/bh-more-details-ui/

hursey013:
@davemcorwin yes I have another branch going to implement a more mobile friendly view for the results page as well - will open another PR for that one today.

Issue #352, opened 2019-09-24
Create accessible tooltip component


hursey013:
Several tasks rely on a tooltip component being in place. We need to spend some time evaluating and testing a solution that we can integrate in with our site since that functionality does not currently exist.

hursey013:
@allisonnorman for prioritization

hursey013:
Moving this to the backlog for the time being since we've implemented a different solution for MVP.

hursey013:
Some notes from the USAgov team:

> The script for our tooltips can be found on GitHub here:
> https://github.com/usagov/webtemplates/blob/dev/assets/js/tooltip.js
> 
> The basic structure for the HTML is like this:
> <span class="tooltip-left dottedUnderline" role="tooltip" tabindex="0">
>     Tooltip Text
>     <img class="tooltip-icon-beta" src="/images/Icon_Tooltip.png" alt="tooltip" aria-hidden="true">
>     <span class="tooltip-left-text">This is the descriptive text that gets displayed when the tooltip is active.</span>
> </span>
> 
> The styles we use for the tooltip can be found here:
> https://github.com/usagov/webtemplates/blob/dev/sass/downlevel/_tooltip.scss
> The styles might be the trickiest part because they were developed alongside our existing styles so they may require a few tweaks to apply them to another site.

PR #393, opened 2019-09-27
(WIP) Implement accessible tooltip

hursey013:
Resolves https://github.com/18F/samhsa-prototype/issues/352


hursey013:
Closing this for the time being until we have time to think through the best way to implement this in an a11y/mobile friendly way.

Issue #365, opened 2019-09-25
Text for tooltip/integrated content on results


amandacostello:
Tooltip text for filter titles

TREATMENT TYPE
There are many types of and places that treat substance use disorder; your recovery may include more than one of these. [See more details about these types of treatment](https://findtreatmentbeta.18f.gov/content/treatment-options/types-of-treatment)

PAYMENT OPTIONS
Understanding paying for treatment can be hard and confusing, but don’t let this stop you from finding help.
- [If you have insurance](https://findtreatmentbeta.18f.gov/content/paying-for-treatment/if-you-have-insurance)
- [If you don’t have insurance](https://findtreatmentbeta.18f.gov/content/paying-for-treatment/if-you-dont-have-insurance)

MEDICATION-ASSISTED TREATMENT (MAT)
Medication-assisted treatment (MAT) is often a successful treatment for people who are addicted to opioids or alcohol. This treatment uses FDA-approved medication together with counseling and behavioral therapies.
[Learn more about the medications approved for MAT] (https://findtreatmentbeta.18f.gov/content/treatment-options/medications-used-in-treatment) and how they can help with recovery.

BEFORE YOU VISIT (starting state collapsed, info-circle icon or other thing to expand box)
- Call a facility before you visit to make sure they have the services you need. [What to expect when you call. ](https://findtreatmentbeta.18f.gov/content/treatment-options/calling-a-facility)
- Not sure what you need? Learn more about different types of treatment. 

outside of/below box but before search results
_All facilities are licensed by their states, and provide assessments._

_Originally posted by @amandacostello in https://github.com/18F/samhsa-prototype/issues/218#issuecomment-534715253_

amandacostello:
In case we want to do-away with the tooltips, here's another version, links to live under the headers.

TREATMENT TYPE
[Details about treatment types](https://findtreatmentbeta.18f.gov/content/treatment-options/types-of-treatment)

PAYMENT OPTIONS
[Not sure how to pay?](https://findtreatmentbeta.18f.gov/content/paying-for-treatment/understanding-the-cost-of-treatment)

MEDICATION-ASSISTED TREATMENT (MAT)
[Details about FDA-approved medications](https://findtreatmentbeta.18f.gov/content/treatment-options/medications-used-in-treatment)

BEFORE YOU VISIT (starting state open, info-circle icon or other thing to expand box)

[...]

PR #403, opened 2019-09-30
Add integrated content links

hursey013:
Resolves https://github.com/18F/samhsa-prototype/issues/365
...with the exception of finding a home for `All facilities are licensed by their states, and provide assessments`.


src/components/Results/ResultsWarning.js
```diff
@@ -3,28 +3,33 @@ import tw from 'tailwind.macro';
...
-        css={tw`underline text-gray-darkest`}
-      >
-        Learn more about different types of treatment
-      </Link>
-      . All facilities are licensed by their states, and provide assessments.
-    </p>
+    <ul css={tw`pl-4 ml-4 list-outside list-disc`}>
+      <li css={tw`mb-2`}>
+        Call a facility before you visit to make sure they have the services you
+        need. What to expect when you call.{' '}
```
allisonnorman: What to expect when you call is duplicated. I think we should just leave the hyperlinked sentence.

src/components/Results/ResultsWarning.js
```diff
@@ -3,28 +3,33 @@ import tw from 'tailwind.macro';
...
+        </Link>
+        .
+      </li>
+      <li>
+        Not sure what you need?{' '}
+        <Link
+          to="/content/treatment-options/types-of-treatment"
+          css={tw`underline text-gray-darkest`}
+        >
+          Learn more about different types of treatment
```
allisonnorman: @amandacostello - I think this needs a 'the' - Learn more about THE different types of treatment.

src/components/Results/ResultsWarning.js
```diff
@@ -3,28 +3,33 @@ import tw from 'tailwind.macro';
 import 'styled-components/macro';
 import { Link } from 'react-router-dom';
 
-import Warning from './ResultsWarning';
+import Warning from '../Warning';
 
 const ResultsWarning = () => (
   <Warning heading="Before you call">
```
allisonnorman: @amandacostello - I think we talked about switching this up. It probably should be "before you visit" or "before you go" as the warning is more around calling before you visit.

src/components/Results/ResultsWarning.js
allisonnorman: Or, "learn more about treatment options"

src/components/Results/ResultsWarning.js
hursey013: Fixed 👍

allisonnorman:
Made the content updates and resolved the change request. Ready to merge, @hursey013

hursey013:
@allisonnorman did we want to find a spot for `All facilities are licensed by their states, and provide assessments.`?

PR #404, opened 2019-10-01
Prevent duplicate searches

### LINKED ISSUE #397: Prevent duplicate searches with same parameters
Right now on the results page, once you entered a location, you can continue to press enter on the location field and the search will re-run. We should prevent this from happening to save unnecessary API calls.

davemcorwin:
Resolves #397 

This is not my ideal solution, but it seems to work. Ideally, when the form submits successfully, the form is reset to a "pristine" state such that we can just check the `touched` or `dirty` values to determine if any values changed and omit submitting if so. Unfortunately, there are some issues with how we are using `redux-form` that is preventing this solution but are out of the scope of this issue.

If it is worth the time, someone can go through in detail and create some issues on the backlog to refactor some of this.

### DIFF
--- a/src/components/Form/FormFilters.js
+++ b/src/components/Form/FormFilters.js
@@ -29,16 +29,6 @@ export class FormFilters extends Component {
     }
   }
 
-  handleSubmit = submitEvent => {
-    const { handleSubmit, location } = this.props;
-
-    if (!(location || {}).latLng) {
-      return submitEvent.preventDefault();
-    }
-
-    handleSubmit(submitEvent);
-  };
-
   renderSubmitButton = () => {
     const { loading, location, recordCount, toggleFilters } = this.props;
 
@@ -65,7 +55,7 @@ export class FormFilters extends Component {
   };
 
   render() {
-    const { isDesktop, languages, resetForm } = this.props;
+    const { handleSubmit, isDesktop, languages, resetForm } = this.props;
 
     return (
       <div css={tw`bg-teal-lighter rounded shadow border border-gray-light`}>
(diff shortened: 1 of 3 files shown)


src/components/Form/FormFilters.js
```diff
@@ -29,16 +29,6 @@ export class FormFilters extends Component {
     }
   }
 
-  handleSubmit = submitEvent => {
```
davemcorwin: This was NOT being called...

src/components/Form/FormFilters.js
```diff
@@ -79,7 +69,7 @@ export class FormFilters extends Component {
             </Button>
           </div>
         )}
-        <form onSubmit={this.handleSubmit}>
+        <form onSubmit={handleSubmit}>
```
davemcorwin: this is the idiomatic way of using [redux-form](https://redux-form.com/8.2.2/docs/api/props.md/#-code-handlesubmit-eventorsubmit-function-code-)

src/components/Form/FormFilters.js
hursey013: Good catch, this was leftover from when when the mobile view had a submit button to fire off the search - now it is using the same `onChange` event that the desktop version uses.

src/components/Results/index.js
```diff
@@ -49,11 +50,16 @@ export class Results extends Component {
     });
   };
 
+  previousValues = null;
```
hursey013: When initiating a search from the homepage this will be set to null even though there will be previous values present - however the added complexity of trying to check for that scenario is probably not worth the effort at this point. Right now the only drawback to not checking is that the search can be re-run with the same values one additional time before `previousValues` is set properly.

hursey013:
@davemcorwin would also like to hear about ways to improve our `redux-form` implementation in case there are built in validation tools we could be using (and for my own understanding!)

src/components/Results/index.js
davemcorwin: Good point. We could throw them in the Redux store, but I am ok with this as is for now.

Issue #364, opened 2019-09-25
Ask ET about which facility fields are optional and required


hursey013:
I am making some assumptions about what data will be available for each facility (like address, phone number, etc) but it would be good to know which of these data points are required and optional so we can better handle instances when the data is not available.

allisonnorman:
@BrooklynHopkins - Would love to chat with you about this on Thursday. 

We need to think about how should we handle facilities that don't have an address, services, payment option details, etc? Brian has already put some things in place to hide phone number and website if not present, but the absence of some of the other data points could look really weird with our design (empty payment option box on detail page, empty map, completely empty bottom section on details page). It's probably worth quantifying how many facilities don't have this but either way, we probably want to account for this. Let's chat more about it tomorrow.

allisonnorman:
Here's what Brooklyn is checking on:
- # of facilities without payment options: 84. Workaround: could we display a message in the payment box that says something like "Check with the facility for payment options" if no payment options are returned? 
- # of facilities without location: 0. Brooklyn said this is not possible.
- [ ] # of facilities without 0 service codes: 0. @hursey013 said he'll follow up with a couple of examples
- [ ] # of facilities without treatment types (type of care or service setting) - @BrooklynHopkins - can you check on this?

hursey013:
Yep this makes sense, looks like missing  payment options is the only thing that we’ll need to account for 👍🏼

hursey013:
Also the cards already account for missing data, it’ll just omit it if there is nothing there.

PR #405, opened 2019-10-01
Show message when no payment options are available

hursey013:
Resolves https://github.com/18F/samhsa-prototype/issues/364

Shows message on details page when no payment options are available:
Screen Shot 2019-10-01 at 12 51 24 PM

### DIFF
--- a/src/components/Details/DetailsPayment.js
+++ b/src/components/Details/DetailsPayment.js
@@ -19,15 +19,21 @@ const DetailsPayment = props => {
               Payment, insurance, or funding accepted
             </h2>
             <ul>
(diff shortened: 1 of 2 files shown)


davemcorwin:
On the results page, is it just blank after *Payment Accepted*? Should we add the same text there?

Screen Shot 2019-10-01 at 11 00 00 AM

hursey013:
Yes - on the card the payment section is omitted:
Screen Shot 2019-10-01 at 2 04 22 PM

@allisonnorman @amandacostello any preference on whether to show a message here as well?

allisonnorman:
I think we can just omit it on the facility card.

Issue #298, opened 2019-09-10
Missing skip navigation link


iamjolly:
### Problem Description
The header of the page/site offers no way to bypass repeated blocks of information, aka Skip Navigation/Skip to Main Content link. Keyboard-only and screen reader users may need to go through the repeated links each time a page loads. Providing a skip link (bypass block) helps users jump to the main content more easily, and it satisfies the [WCAG 2.4.1 - Bypass Blocks (Level A)](https://www.w3.org/TR/2008/REC-WCAG20-20081211/#navigation-mechanisms-skip) requirement for Section 508 compliance.

### Solution
* [ ] Add a skip to main content link as the first focusable element in the DOM, and link it to the `main` element on the page (you'll want to give `main` an `id` attribute).
* [ ] Make the skip to main content link visually hidden when not focused.
* [ ] Show (un-hide) the element visually when it has `focus`.

### Examples/Articles
* [WebAIM - Skip Navigation Links](https://webaim.org/techniques/skipnav/)
* [Knowbility - Skip links are important](https://knowbility.org/blog/2019/skip-links/)
* [Knowbility - Skip Links Design Showcase](https://knowbility.org/blog/2019/skip-links-gallery/)
* [USWDS - Components - Header](https://designsystem.digital.gov/components/header/) - Look under "Header accessibility" for guidance about skip links.

PR #407, opened 2019-10-01
Add skip link

hursey013:
Resolves https://github.com/18F/samhsa-prototype/issues/298

Preview:
Screen Shot 2019-10-01 at 3 13 09 PM

To test, use tab to navigate webpage. Pressing the skip link will redirect the focus of the page to `<main />`.


davemcorwin:
@iamjolly May have a better sense of this, but should this be an `a` tag that points to `#main` (assuming the `id` is added) instead of a `button` tag that uses JS to set the focus?

davemcorwin:
@Jkrzy do you have thoughts on this?

hursey013:
@davemcorwin I can update to use `#main` instead.

hursey013:
@davemcorwin piggybacking off your `ReturnToTop` component, this updated the skip nav to use a normal anchor link while also focusing the content.

hursey013:
@davemcorwin holding off on merging this as well because I think it will conflict with https://github.com/18F/samhsa-prototype/pull/411

Issue #212, opened 2019-08-07
Build architecture to support localization


allisonnorman:
This is lower priority on our backlog, but would love to discuss this more with @hursey013 and @Jkrzy given the timeline of this project. Creating an issue so that we can keep track of this.

Jkrzy:
Given that _most_ of our content is already isolated in a few places we're a bit ahead of the curve on this one. Next steps here are a feature branch with a foundational implementation of https://github.com/formatjs/react-intl.

Once it's working as expected on that branch we can began slowly migrating content and building out i18n capability.

hursey013:
There's also https://github.com/i18next/react-i18next, although it looks like `react-intl` is more widely used, but might be worth checking out the differences.

PR #411, opened 2019-10-02
Add a11y navigation features

### LINKED ISSUE #314: Find treatment > Results transition == problems with keyboard-only and screen readers
### Problem Description
When searching for a treatment facility using a keyboard only and/or a screen reader (Safari+VoiceOver on MacOS), the transition to the results does not give information to people with disabilities abut the page state/content changes. Specifically, the following problems are noted:
* When results are loading, an animation is shown but no announcement is made that something is happening.
* As the results load, it appears as if it is a new page (and the URL updates), but no announcement is made as to the page title or that results have loaded. 
* Focus is not managed with the transition to the new screen/page of results. Using Safari + VoiceOver, the focus is somewhere in the middle of the filters, when it would be expected to be either at the top of the page or on the `h1` for the results ("Showing XXX facilities in Selected City, ST, 99999"). 

#### Quicktime movie of what Safari+VoiceOver on MacOS reads out for this interaction:
[samhsa-vo-find-results-transition.mov.zip](https://github.com/18F/samhsa-prototype/files/3607616/samhsa-vo-find-results-transition.mov.zip)

### Recommended Solution
* [ ] For pages that load results and show an "in progress" or loading animation/spinner, ensure that either focus is set to that element or that an `aria-live` announcement is made while that element is visible.
* [ ] Update users about the new "page" title when those are changed by the app.
* [ ] Set focus to the most appropriate location on the new "page" or screen whenever possible. When not possible, set focus to the top of the page and provide a skip link to allow users to jump to the main content of the new page/screen.

### Related Articles/Examples
* [Accessible Routing in React](https://timwright.org/blog/2019/03/23/accessible-routing-in-react/)
* [Accessible React Router navigation with ARIA Live Regions and Redux.](https://almerosteyn.com/2017/03/accessible-react-navigation)
* [Creating accessible React apps](http://simplyaccessible.com/article/react-a11y/)

davemcorwin:
Resolves #314 

### On initial page load
1. Set the page title

### On page transitions EXCEPT initial page load
1. Set the page title
2. Set the new page title as an `aria-live` message for SRs
3. Set focus to the first `h1` OR if none, the first `h2`
4. Scroll to the top of the page

### On the Results page AFTER loading as finished
1. Set focus to the first `h1` OR if none, the first `h2`
2. Scroll to the element with focus
_Here I am NOT explicitly making a SR announcement since when the `h1`/`h2` receives focus, the text content is read.

## Future Work
This is step 1 of these updates. The next steps will be to enhance the search experience by 
1. Providing a more detailed sr announcement that mimics the content shown in the header (ex. "Showing 43 results")
2. adding an sr/on-focus only "skip to results" link at the top of the filters that will navigate to the results header.
3. Add an sr announcement when the results are loading, only after a delay of x seconds

## Implementation
This is implemented by piggy-backing off of the callback provided by `react-helmet` that is run on transition. Since the "scroll to top" functionality is implemented here, I removed the extra component that was previously responsible for it. It allows us some control since a subsequent use of `Helmet` can override any of the previous attributes.


src/components/AppHelmet.test.js
```diff
@@ -0,0 +1,39 @@
...
+  setSRMessage: jest.fn()
+};
+
+describe('App Helmet component', () => {
+  it('should return a `Helmet` component', () => {
+    const wrapper = shallow(<AppHelmet {...testProps} />);
+    expect(wrapper.is('HelmetWrapper')).toBe(true);
+  });
+
+  // it('should call `setSRMessage` on navigation', () => {
```
davemcorwin: still looking into how to test these pieces

src/components/Content.js
```diff
@@ -143,56 +143,42 @@ export class Content extends Component {
...
-            <StyledPage>
-              {this.renderMain(section)}
-              {this.renderSideBar()}
-            </StyledPage>
-          </div>
-          <ReturnToTop />
-        </div>
-      );
-    } else {
+    if (!section) {
```
davemcorwin: refactored to remove the extra render that results from the redirect

src/components/Form/FormFilters.js
```diff
@@ -46,7 +46,7 @@ export class FormFilters extends Component {
   };
 
   render() {
-    const { handleSubmit, isDesktop, languages, resetForm } = this.props;
+    const { handleSubmit, isDesktop, resetForm } = this.props;
```
davemcorwin: refactor to remove unused variable resulting from a previous PR

src/components/Results/index.js
```diff
@@ -174,4 +186,13 @@ const mapStateToProps = state => {
   };
 };
 
-export default connect(mapStateToProps)(Results);
+const mapDispatchToProps = {
```
davemcorwin: include some refactoring to be consistent on how action creators are dispatched

src/components/ScrollToTop.js
```diff
@@ -1,27 +0,0 @@
-import { Component } from 'react';
```
davemcorwin: This functionality is now handled in `AppHelmet`

src/reducers/ui.js
```diff
@@ -13,6 +18,17 @@ export default function ui(state = initialState, action) {
...
+    case SET_SR_MESSAGE:
+      return {
+        ...state,
+        srMessage: action.message
+      };
+    case RECEIVE_FACILITIES_FAILURE:
+    case RECEIVE_FACILITIES_SUCCESS:
+      return {
+        ...state,
+        srMessage: 'New results are available.'
```
davemcorwin: Not sure if there is a better place or value for this string

davemcorwin:
currently looking at the failing tests...

src/components/Results/index.js
```diff
@@ -36,17 +37,18 @@ export class Results extends Component {
     }
 
     this.clearResultsIfNoLocation();
+    this.focusTarget.current && this.focusTarget.current.focus();
```
davemcorwin: This check was required to make the tests pass. I attempted to use `mount` instead of `shallow` but then the `context` was not available. I don't know know why and couldn't find an answer to this specific issue on the web, solutions to enable child components access to `context` are not relevant and don't help, neither does providing a "wrapper" component with the context.

src/components/AppHelmet.js
```diff
@@ -0,0 +1,29 @@
...
+import { setSRMessage } from '../actions/ui';
+
+export const AppHelmet = ({ focusTarget, setSRMessage }) => (
+  <Helmet
+    titleTemplate={`%s | ${process.env.REACT_APP_SITE_TITLE}`}
+    defaultTitle={process.env.REACT_APP_SITE_TITLE}
+    onChangeClientState={newState => {
+      setSRMessage(newState.title);
+      focusTarget.current.focus();
+      window.scrollTo(0, 0);
```
hursey013: Looks this is getting called onMount and moving the focus to `main` - I wonder if having a conditional similar to the old `ScrollToTop` component could help: `this.props.location.pathname !== prevProps.location.pathname` to make sure it's only getting set on route changes?

hursey013:
Would it make sense to make setting the focus to `main opt-in? On a normal page transition, I think it would make sense to move the focus to the top of the page like you would see in a non-SPA website (especially since we will have a skip nav), but in the case of the results page, where new records are being shown when filters are changed, perhaps move the focus to `main`?

src/components/Content.js
hursey013: I like the looks of this, but functionally it's slightly different than what I think the original intent was. In the live code, clicking one of the top nav links redirects the user to the first child page of that section, including the proper URL. In this refactor it displays the content of the child page, but it does not update the url which in turn doesn't highlight the proper page in the nav:

Screen Shot 2019-10-02 at 5 10 19 PM

I definitely think the current code needs to be refactored, but I wonder if we should split this off into a separate PR so we maybe solve this at a higher level from a content/IA perspective?

davemcorwin:
@hursey013 Just checked in with @iamjolly and you're spot on. I'm going to update to set the focus on the top of the page. Once the skip-nav pr is merged, the focus will be set BEFORE it.

src/components/Content.js
davemcorwin: Reverted, the re-render is not causing issues.

src/components/AppHelmet.js
davemcorwin: Where is this an issue?

davemcorwin:
@hursey013 @iamjolly ready for another look. I'm not seeing that issue we observed in Safari anymore either @iamjolly

src/components/App.js
```diff
@@ -19,18 +18,25 @@ import Details from './Details';
...
+  topRef = React.createRef();
+
   render() {
     return (
-      <ScreenContext.Provider value={this.props}>
-        <Helmet
-          titleTemplate={`%s | ${process.env.REACT_APP_SITE_TITLE}`}
-          defaultTitle={process.env.REACT_APP_SITE_TITLE}
-        />
+      <ScreenContext.Provider value={{ ...this.props, topRef: this.topRef }}>
```
hursey013: I've wondered when else it'd be useful to use context - this is great!

src/components/SRAnnouncements.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import { connect } from 'react-redux';
+import { PropTypes } from 'prop-types';
+
+export const SRAnnouncements = ({ srMessage }) => (
+  <div className="sr-only" role="status" aria-live="polite" aria-atomic="true">
+    {srMessage ? <span>{srMessage}</span> : ''}
```
hursey013: Just curious if the `<span>` is required?

src/components/SkipNav.js
```diff
@@ -14,8 +14,8 @@ const StyledLink = styled(HashLink)`
   }
 `;
 
-const SkipNav = () => {
-  return <StyledLink to="#main">Skip to main content</StyledLink>;
+const SkipNav = ({ to }) => {
+  return <StyledLink to={to}>Skip to main content</StyledLink>;
```
hursey013: This visually moves the page to the correct place, but does not trigger Helmet's `onChangeClientState` so the focus is never set and remains on the skipNav button.

src/components/AppHelmet.js
```diff
@@ -0,0 +1,30 @@
...
+import { connect } from 'react-redux';
+import PropTypes from 'prop-types';
+import { setSRMessage } from '../actions/ui';
+
+export const AppHelmet = ({ focusTarget, setSRMessage }) => (
+  <Helmet
+    titleTemplate={`%s | ${process.env.REACT_APP_SITE_TITLE}`}
+    defaultTitle={process.env.REACT_APP_SITE_TITLE}
+    onChangeClientState={newState => {
+      setSRMessage(newState.title);
```
hursey013: What do you think of stripping off the ` | TreatmentFinder` text that gets appended to the page title? Seems unnecessary for screen readers.

```suggestion
      setSRMessage(newState.title.replace(` | ${process.env.REACT_APP_SITE_TITLE}`, ''));
```

src/components/Results/index.js
```diff
@@ -11,6 +11,7 @@ import {
   destroyFacilities,
   handleReceiveFacilities
 } from '../../actions/facilities';
+import { setSRMessage } from '../../actions/ui';
```
hursey013: Is `setSRMessage()` ever used in this component?

hursey013:
This is looking great - left a couple comments/questions, but looks like it's getting close.

The only other thing I noticed when testing it with VoiceOver was that when a page is initially loaded (homepage, /results, etc) the focus is set to `top`, but since `top` wraps the entire page VoiceOver begins reading _all_ of the content on the page. This doesn't appear to happen when transitioning between pages though.

I don't know best practice here, perhaps @iamjolly can weigh in, but I wonder if it's be ok for `top` to just be an empty div or something at the top of the page rather than a wrapper div, so that the focus is moved to the right place without VO trying to read everything?

src/reducers/ui.js
hursey013: I think we've definitely crossed over into benefiting from a proper i18n/localization implementation, we have an issue here https://github.com/18F/samhsa-prototype/issues/212, but for now I think leaving them in here is fine.

src/components/AppHelmet.js
hursey013: I was, probably inaccurately, trying to solve for the issue described in https://github.com/18F/samhsa-prototype/pull/411#pullrequestreview-296467378 and https://github.com/18F/samhsa-prototype/pull/411#issuecomment-537986456. I think we need some way of preventing this callback from happening on the initial/fresh load of a page.

src/components/SRAnnouncements.js
iamjolly: `span` isn't needed from an a11y perspective there. Everything within the `div` with `aria-live` on it should get announced, regardless of the inner elements.

src/components/SkipNav.js
iamjolly: Nice catch @hursey013. I see the same behavior.

src/components/AppHelmet.js
iamjolly: I like having the `| TreatmentFinder` appended to the announcement, just like it is in the `title` element for context, but this is one that could go either way and still be usable.

src/components/SRAnnouncements.js
davemcorwin: huh, yeah, for some reason I thought an extra wrapping element was required, but reading back through the examples i followed, it's never mentioned, removing...

src/components/Results/index.js
davemcorwin: nope, leftover from previous changes, removing...

src/components/SkipNav.js
davemcorwin: ugh, when clicking on a link is no longer clicking on a link...REACT ROUTER!!!!

davemcorwin: OK, updated this and `ReturnToTop` to set focus on the targeted element.

iamjolly:
I'm getting some strange-ish behavior when using VoiceOver with Safari on the pages here. When I use tab like a sighted keyboard-only user, everything seems to work *perfectly*. However, with Safari and VO, when I read to the different links (using `ctrl` + `option` + `arrow`) and activate one any of the links (using `ctrl` + `option` + `space`), the focus/cursor seems to stay on the link and doesn't move to the top of the page. 

It seems like (because I didn't look at the code changes closely here) that the setting of focus is firing only when `enter` is pressed on the links and not when all the possible execution events on the anchor/link are fired (mouse click, etc.). Is that what's happening here?

Video of me rambling on and demo-ing this with Chrome and Safari with/without VoiceOver. 
[zoom_1.mp4.zip](https://github.com/18F/samhsa-prototype/files/3687350/zoom_1.mp4.zip)

iamjolly:
> I don't know best practice here, perhaps @iamjolly can weigh in, but I wonder if it's be ok for top to just be an empty div or something at the top of the page rather than a wrapper div, so that the focus is moved to the right place without VO trying to read everything?

I think it's OK for the contents to read out automatically if focus is set to the container. That is pretty standard behavior for SRs. Another option, if it's possible to try, is setting the focus to the `body` element of the page. That's essentially what happens on full page loads (or at least what I can surmise with using `document.activeElement` on fresh page loads on other sites).

davemcorwin:
@iamjolly It appears that Voice Over will only "focus" on certain elements. When I change the focus target to an `h1` it moves the cursor to the expected location, but for `div`, `main`, `header` it does not work. It also does not work if I clear the focus entirely (resulting in `document.activeElement === body`)

It's possible that this behavior is specific to the screen reader, but i just tried ChromeVox and I think the same thing is happening.

This seems wonky to me because we actually don't always want to focus on the `h1` when navigation throughout the site. I'm not sure how to solve this.

I can't find anything on the interwebs that addresses this issue, every solution for SPAs and screen readers just mentions the `aria-live` region and setting the focus. 

AAAAND, i just encountered another strangeness. I am currently just clearing focus altogether, when I select a sidenav link with the screen reader it reads the "skip nav" content but the cursor doesn't move...

I'm at a loss...

iamjolly:
@davemcorwin - Let's find time Monday or Tuesday to meet and review this more. VoiceOver shouldn't have different points of focus than the underlying browser, so it may be worth going through it together to see what is happening and dig into what may be possible. Hang in there. We'll get this settled and working well.

davemcorwin:
I have found discussions on the interwebs that do acknowledge the difference btwn screen reader focus and keyboard focus (ex. https://webaim.org/discussion/mail_thread?thread=8526). That said, Brian suggested setting focus on the “skip nav” link which seems to “work”, is this s reasonable solution??

davemcorwin:
The downside is that now clicking on a link makes the "skip nav" link visible which is a bit weird for non-keyboard navigators...

hursey013:
Ugh, well that would make sense, but yeah was not considering that 🙁

davemcorwin:
I'm confused. The "skip nav" link seems to work in all situations (click, <enter>, VO). When I used the same functionality for the links on the Content page, it DOES NOT work!!!  @iamjolly let's touch base next week.

iamjolly:
I really like the focus getting set to the `h1` on the pages. Seems to work properly in Safari with VoiceOver now.

iamjolly:
> I'm not hearing the SR announcements in the `aria-live` region anymore with the page's titles. This is with Safari and VoiceOver on MacOS. I'll check Firefox and JAWS on Windows now.

Yep. The functionality is pretty much perfect at this point for setting the focus when new info is loaded on the page. Having the focus set to the headings is nice. What's missing is the `aria-live` for announcing the page title, as sometimes the `h1` doesn't exactly match the page title, so having that announced along with focus set on the `h1` or `h2` would put a bow on this and deliver it wonderfully @davemcorwin. Thanks for the work on this one!

iamjolly:
Tested with Firefox and Chrome using JAWS on Windows 10. Also tested with Safari using VoiceOver on MacOS.

davemcorwin:
@iamjolly Thank you for catching that! There was a bug there that has been addressed. However, I'm not sure we are able to achieve the desired result.

When I test, there seems to be a conflict between the announcement of the new title and reading the h1. If I leave it as is, the announcement of the new title is superseded by reading the content of the h1 so I don't hear it. If i change `aria-live` to `assertive` I hear the announcement of the new title, but the content of the h1 is no longer read, though it does receive focus. I've tried adding a small delay to setting the focus to the h1 but it does not seem to help.

What is the desired behavior here? Is it to announce the new title, then read the h1? If we want to hear the title, then hear "you are on a heading level 1" that the user can select to hear, then `aria-live="assertive"` may be the way to go. I am also aware that actual users may listen at a different speed, so its possible this is an artifact of that...

iamjolly:
@davemcorwin - thanks for thinking through all this. I have some comments based on your dilemma—hopefully this will get things in a resolved and merge-worthy state without much else needed. :)

> If I leave it as is, the announcement of the new title is superseded by reading the content of the h1 so I don't hear it.

Right now, I get the `h1` announced and then the `aria-live` spoken with the page title. This could be due to the rate/speed of how I have VoiceOver set. I think this behavior is totally fine. If you're not hearing the `h1` but the `aria-live` page title _is_ getting announced, I think that is also OK. 

> What is the desired behavior here? Is it to announce the new title, then read the h1? If we want to hear the title, then hear "you are on a heading level 1" that the user can select to hear, then `aria-live="assertive"` may be the way to go.

I like the idea of having `aria-live="assertive"` set in this case. The most important thing is announcing the page title, and with the focus set to the `h1`, users are in a good place to continue with confidence that the page content has changed (due to the SR Announcement) and will get the info they need. If we can go with that, it will reduce the amount of audio slightly while keeping clarity. 

Thanks for digging in and working through this one! It's in great shape because of that.

davemcorwin:
thanks 💯 @iamjolly !!! Updated to use `aria-live="assertive"`

PR #429, opened 2019-10-02
Add Helpline to mobile footer banner

### LINKED ISSUE #418: Helpline location on mobile.
## Current State
- On mobile the only place the National Helpline is seen is when you open the nav menu. 
- At the bottom of each page on mobile is The Suicide Prevention hotline 

## Question 
- Should we surface the the treatment Helpline more? 

## Ideas
- On mobile - add the National Helpline directly above the Suicide Prevention hotline info, either in the same box or in a separate one. 
- OR: on mobile - remove the Suicide Prevention hotline and in its place add the National Helpline.

allisonnorman:
This is my attempt at tackling #418. 
- Added Helpline number to Suicide Prevention/911 footer banner only on mobile

Ben brought this up. I know we made the call to put Helpline under the hamburger menu but it feels odd to show Suicide Prevention Lifeline and not the Helpline. I think we should keep Helpline in the hamburger menu on mobile either way.

We can opt not to move forward with this but wanted to flex my flex/tw skills and try it. 

@bpdesigns - is this what you were hoping for? As expected, it adds more height to an already large footer banner. Worth it, or not? 

Resolves #418


src/components/Footer.js
```diff
@@ -161,23 +161,34 @@ const Footer = () => {
...
-          at{' '}
-          <OutboundLink
-            eventLabel="Suicide prevention hotline link from footer"
-            to="tel:1-800-273-8255"
-            css={tw`underline whitespace-no-wrap`}
-          >
-            1-800-273-8255
-          </OutboundLink>{' '}
-          | In emergencies, dial 911.
+          <p css={tw`sm:hidden`}>
```
hursey013: The mobile menu begins to appear at the `md` breakpoint - so in order to keep this in sync with that, we should use:
```suggestion
          <p css={tw`md:hidden`}>
```
I think visually it might make be good to add some space between the helpline and suicide prevention paragraphs so we could also add a `mb-4` (margin bottom of 4 units).

src/components/Footer.js
```diff
@@ -161,23 +161,34 @@ const Footer = () => {
...
+            For help finding treatment, call <br />
+            <OutboundLink
+              eventLabel="Helpline link from mobile footer"
+              to="tel:1-800-663-4357"
+            >
```
hursey013: For consitency with the other phone number, lets underline it, and also make sure that the number itself never splits onto two lines: 
```
css={tw`underline whitespace-no-wrap`}
```

This will allow you to remove the hardcoded `<br />` on line 165 as well.

src/components/Footer.js
```diff
@@ -161,23 +161,34 @@ const Footer = () => {
...
+              1-800-662-HELP (4357).
+            </OutboundLink>{' '}
```
hursey013: The extra space here is not needed:
```suggestion
            </OutboundLink>
```

allisonnorman:
As there's been a couple updates here as well, flagging that it's ready for approval, @hursey013. Thanks for your help!

src/components/Footer.js
```diff
@@ -161,23 +161,35 @@ const Footer = () => {
...
-          </OutboundLink>{' '}
-          | In emergencies, dial 911.
+          <p css={tw`lg:hidden mb-4`}>
+            For help finding treatment, call{' '}
+            <OutboundLink
+              eventLabel="Helpline link from mobile footer"
+              to="tel:1-800-663-4357"
+              css={tw`underline whitespace-no-wrap`}
+            >
+              1-800-662-HELP (4357).
```
hursey013: Very small nitpick, but we should move the period outside of the hyperlink for the benefit of screenreaders.

Issue #431, opened 2019-10-03
Add meta description to details page


hursey013:
To support SEO/social sharing

amandacostello:
Generic `meta` content for a details page:

_Treatment facility details, including what services it provides, who it serves, and what payments are accepted._

PR #436, opened 2019-10-03
Add meta description to details

hursey013:
Resolves https://github.com/18F/samhsa-prototype/issues/431


allisonnorman:
I don't know if I can test this until it's merged but 👍 and can check it after!

Issue #441, opened 2019-10-04
remove site from federalist staging


amirbey:
@hursey013  to:
- [x] remove site from federalistapp-staging.18f.gov
- [x] remove federalistapp-staging webhooks in GitHub - https://github.com/18F/samhsa-prototype/settings/hooks

amirbey:
added to federalist staging only for testing of react-snap .... should be removed to prevent staging from occasionally sticking PR checks

davemcorwin:
Removed from Federalist Staging

allisonnorman:
@davemcorwin - Can we close the issue?

davemcorwin:
@allisonnorman yes!

PR #443, opened 2019-10-04
Implement prerendering with react-snap

hursey013:
We have meta tags on all pages now, and while that'll work for search engines like Google, social networks like Facebook and Twitter are not able to crawl single page applications. I tested out meta tags with Twitter (https://cards-dev.twitter.com/validator) and it did not pick up any of the meta tags.

This implements https://github.com/stereobooster/react-snap which prerenders the application in HTML. It relies on headless chrome / puppeteer, which was recently installed in Federalist Staging (https://federalistapp-staging.18f.gov/). I was able to get it working but have run into an issue with https://github.com/renatorib/react-sizes use of Context API. As you can see here, the pieces of JSX that are meant to be conditionally rendered are all rendered at the same time so you get the mobile and desktop elements displaying: https://cg-c90c8909-4f4d-4660-b5a5-15cf92597007.app.cloud.gov/preview/18f/samhsa-prototype/bh-react-snap/ - Once you click around the javascript is run properly, so the proper elements are displayed.

To complete this, it may be necessary to remove or reduce the usage of react-sizes and use normal CSS media queries instead.


hursey013:
Related: https://github.com/18F/samhsa-prototype/issues/441

hursey013:
This is so close... but few additional tasks:
- headless chrome needs to be moved into production on Federalist before this will work - it is currently only on Federalist staging. Amir mentioned he will be able to do this next week, mon/tues
- react-snap is having some conflicts with the react-sizes library that is being used to conditionally show/hide elements based on screen size. i began try to refactor out the react-sizes dependency since that would likely be the best approach, otherwise it might be worth investigating a way to force a rerender to get things displayed correctly onmount.

allisonnorman:
I was able to test this by DMing myself in Slack and it looks good! Posting a screen shot of that for @austinhernandez to take a look too.

Screen Shot 2019-10-10 at 10 50 42 AM

I'm assuming this will show up the same in other social platforms that support previews or should we do more testing?

allisonnorman:
Actually @davemcorwin - a little birdie (Brian, who's still invested in this project - thanks @hursey013 ) mentioned that the large logo icon is related to https://github.com/18F/samhsa-prototype/issues/448. It should show up smaller, like: 

Screen Shot 2019-10-10 at 11 46 28 AM

Is that feasible?

davemcorwin:
This should be regression tested visually, which has to be done locally (until merged to `master`) because of the how Federalist handles redirects and the most important part to check is how the site looks on initial load on mobile and desktop.

To do this locally:
- build the app with `npm run federalist`
- server the app (I use `serve` npm library (`npm i -g serve`) `serve _site`

allisonnorman:
Update: Spent some time doing regression testing locally and am not seeing any issues.

PR #446, opened 2019-10-04
Content updates

amandacostello:
A lot of content has moved around, but no pages have been renamed. I would like the "It's hard to stop on your own" page to be deleted; its content made more sense on another page, so that page is empty now.


allisonnorman:
Good to go though making a note we'll want to do #447 to update the link paths

PR #449, opened 2019-10-04
Used link component on content.js and update internal content links

### LINKED ISSUE #447: Use the link component for intra-site links in content section
We have links to the findtreatmentbeta.18f.gov URL within the content section. This won't work when we switch to FindTreatment.gov (and turn off the beta site).

Note from Brian on this:
so you want to use the `Link` component, so first you need to import it into that file, so up top, on line 4, you'd add: `import { Link } from 'react-router-dom';`
(and then add a line break after it so its separated from all the icon imports below it)
and then anytime you want an internal link you'd use:
```<Link to="/path-to-page">Name of link</Link>```

allisonnorman:
Fixes #447 

@davemcorwin - I followed Brian's directions here but would appreciate your approval to verify if I did it correctly. Let me know if you have any questions! Thanks

### DIFF
--- a/src/utils/content.js
+++ b/src/utils/content.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import tw from 'tailwind.macro';
 import 'styled-components/macro';
+import { Link } from 'react-router-dom';
 
 import { ReactComponent as IconAddiction } from '../images/what-to-expect-addiction.svg';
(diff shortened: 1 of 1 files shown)


BrooklynHopkins:
@allisonnorman @Jkrzy i can merge this?

PR #454, opened 2019-10-10
[WIP] Use nationalmap.gov 

Jkrzy:
Proof of concept for replacing static map generation with a combination of leaflet.js and nationalmap.gov

Herein we leverage the awesome [services](https://viewer.nationalmap.gov/services/) provided by our friends @ https://www.usgs.gov to generate map images, approximating static, for rendering on the facility cards.

Specifically, we're combining the USGS Imagery Topo Base Map WMSServer endpoint with leaflet.js. There are other basemaps and layers to explore.

This is a rough proof of concept that will need refactoring, revision, and enhancement if and when it is to be merged into `master`

Big thanks to @apburnes for the pointers and being a sounding board on this!!


public/index.html
```diff
@@ -29,6 +29,9 @@
       src="//script.crazyegg.com/pages/scripts/0083/6179.js"
       async="async"
     ></script>
+    <link rel="stylesheet" href="https://unpkg.com/leaflet@1.5.1/dist/leaflet.css"
```
apburnes: You could import this directly from the package so you don't need to rely on outside dependency. In the map component, pull in the css with `import 'leaflet/dist/leaflet.css'`

allisonnorman:
@Jkrzy - Created issue #555

Jkrzy:
Closing this for now, conversation and work continued in #555

PR #461, opened 2019-10-11
Removing hyphens

amandacostello:
Removing hyphens/minus-signs for better display on mobile.


allisonnorman:
I see four more "—" that are still rendering badly in Safari. Sent them to @amandacostello.

amandacostello:
Have updated the hyphens situation! Should be good to go @allisonnorman

PR #462, opened 2019-10-11
Report failed search queries

Jkrzy:
Let's report in GA when a visitor attempts but fails to complete a search.


Jkrzy:
Ah great call, thanks for the pointer @davemcorwin!! Changes incoming.

PR #465, opened 2019-10-11
Dc 453 map fail fallback

### LINKED ISSUE #453: Facility card option without map images
@Jkrzy and @davemcorwin are digging into a limitation we have with generating static images from the Google API. There is a limit of 25k per day, which means that we could reach our max with ~2500 searches.

As they investigate potential workarounds, @austinhernandez - could you help by putting sharing a design for a facility card without the static map image, or do you have a previous design that you already created like this? This might need to be used for all cards, or potentially as a fallback option once the limit is reached. 

To be a bit more specific:
- The only asset that we may have issues generating is the static image
- Distance and a link to google maps (or, the distance hyperlinked to google maps) is still on the table.

davemcorwin:
Resolves #453 

This is different than what @austinhernandez designed bc in this case the failure to load the image is happening dynamically. Since we don't know if it fails until it does, I didn't want to move things around which might be jarring to the user. Here, we just replace the google map with an icon of the same size.

The actual logic here shows the placeholder icon first, then replaces it with the google map if it loads successfully.

I am 100% open to suggestions on improving the design and want to see how folks feel about this interaction AND how it looks and feels when deployed.

Screen Shot 2019-10-11 at 3 03 03 PM


Jkrzy:
Implementation looks good, deferring to @austinhernandez and @allisonnorman on visual aspects

davemcorwin:
FYI - working on the mobile version...

davemcorwin:
Screen Shot 2019-10-15 at 11 02 30 AM
Screen Shot 2019-10-15 at 11 02 21 AM

PR #483, opened 2019-10-22
Move focus to What to Expect section when Learn more is clicked

hursey013:
For better keyboard accessibility, this moves the focus to the `What to expect` section when the `Learn more` button is clicked on the homepage.


allisonnorman:
Thanks @hursey013 :) 

@davemcorwin - Can you review when you get the chance?

davemcorwin:
Bummer the CI and Federalist don't run for forks, but we can check this after merge to make sure its ok! Thanks @hursey013 !

Issue #472, opened 2019-10-18
Create production Crazy Egg snapshots


Jkrzy:
Required for: #344 

For release we should establish CE snapshots for each page of the TL.

https://crazyegg.com

Landing and search pages - 6 total to capture mobile and desktop
- [x] Mobile & desktop - landing page
- [ ] Mobile & desktop - results page
- [ ] Mobile & desktop - details page 

Content pages: Snapshots for EACH url - 26 total?
- [ ] Mobile & desktop - /content/treatment-options/types-of-treatment
- [ ] Mobile & desktop - /content/treatment-options/medications-used-in-treatment
- [ ] Mobile & desktop - /content/treatment-options/calling-a-facility
- [ ] Mobile & desktop - /content/treatment-options/what-happens-next
- [ ] Mobile & desktop - /content/treatment-options/who-provides-care
- [ ] Mobile & desktop - /content/paying-for-treatment/understanding-the-cost-of-treatment
- [ ] Mobile & desktop - /content/paying-for-treatment/if-you-have-insurance
- [ ] Mobile & desktop - /content/paying-for-treatment/if-you-dont-have-insurance
- [ ] Mobile & desktop - /content/understanding-addiction/addiction-can-affect-anyone
- [ ] Mobile & desktop - /content/understanding-addiction/recovery-starts-with-asking-for-help
- [ ] Mobile & desktop - /content/understanding-addiction/supporting-recovery-and-self-care
- [ ] Mobile & desktop - /content/understanding-mental-health/serious-mental-illness
- [ ] Mobile & desktop - /content/understanding-mental-health/know-the-warning-signs

Also TODO
- [x] Update `analytics.js` with new `CE_SNAPSHOT_NAME` value [#498]

bpdesigns:
@Jkrzy @BrooklynHopkins This is how I would prioritize CE snapshots considering - 1. what we want to make sure is working now and 2. what would we learn from this that would impact further feature improvements?

**Initial Search** so we understand the initial experience 
- [ ]  Mobile & desktop - landing page
- [ ]  Mobile & desktop - results page

**Content pages linked from filter headers**  so we understand how they react to entering a new page from these links and if/what content it causes them to dig into further
- [ ] TREATMENT TYPE - https://findtreatmentbeta.18f.gov/content/treatment-options/types-of-treatment
[...]

PR #498, opened 2019-10-24
Refactor CE name tracking

Jkrzy:
For #472, track all facility pages using the same snapshot name.

 Also move configuration of CE_SNAPSHOT_NAME tracking title to `.env`


Jkrzy:
Ah good catch! We only need to set this on production. Tweak incoming.

PR #564, opened 2019-11-27
Update README.md

allisonnorman:
- Removed feedback form link (as SAMHSA will no longer have access)
- Fixed the CircleCI bullet point which was rendering with a broken image


allisonnorman:
@Jkrzy - Removed the link from readme. I still see circleci in .circleci/config.yml. Are you removing that too?

Jkrzy:
We'll leave that there for when CircleCI can be enabled 🤞

